PR #5, opened 2021-02-15
Add contribution document

ericzolf:
As the subject says. I also started to add some styling "stolen" from the CER repository.


jscar-hawk (approved):

CONTRIBUTE.adoc
```diff
@@ -0,0 +1,104 @@
...
+  become: false
+
+  tasks:
+
+  - name: say what we all think
+    debug:
+      msg: asciidoctor is {{ my_private_thoughts }}
+----
++
+Even more examples...
```
mophahr, line 103: This seems to be identical to the template above. Does it really need to be in here twice?

CONTRIBUTE.adoc
```diff
@@ -0,0 +1,104 @@
...
+  `asciidoctor` or with `asciidoctor-pdf`.
++
+NOTE: this contribution file is obviously not meant for inclusion in the
+      overall document.
++
+. Each source file has a single title (the line starting with one equal sign)
+  and can be rendered individually (the `leveloffset` is set such that it
+  fits in the overall headings structure when included).
+. The source code is written as readable as possible in its raw form.
+  Lines are kept below 80 characters, indentations are respected.
```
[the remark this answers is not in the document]
ericzolf: Replaced with link to Asciidoc recommended practices and "ventilated" approach (one line for each sentence, one sentence for each line). I discovered it very recently and it makes a lot of sense, have a look.

CONTRIBUTE.adoc
ericzolf: It's because you read the raw code, the idea is to have it once as raw code, once rendered as example, so that people see how it looks. It makes more sense if you look at the rendered page. I've added comments and links accordingly to make this clearer.

PR #14, opened 2021-04-18
Publish good practices to GitHub pages

ericzolf:
There are two commits in this PR:

The first one adds a `docs/index.html` file.

The idea would be to generate `docs/index.html` (and CONTRIBUTE.html) later on automatically through a pipeline. The same pipeline could create also the PDF

For now manual creation will do, so that https://redhat-cop.github.io/automation-good-practices/ offers a rendered version of the good practices.

The 2nd commit changes the recommendations' title into sections. This makes the table of content much more meaningful and renders also better in PDF:
image
Compared to the previous rendering:
image

I kept two separate commits so that you can compare the outputs of the HTML.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # generated files
 *.pdf
 *.html
+!docs/*.html
 
 # temporary files
 .*.swp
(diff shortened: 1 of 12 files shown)


mophahr (approved):
Reviewed and approved in the CoP meeting.

PR #18, opened 2021-06-13
Fix rendering of images in HTML

ericzolf:
Asciidoctor doesn't redirect image links for HTML so we need to copy them
and to place them in one central place.

The instructions have been adapted accordingly and the HTML re-rendered.


jscar-hawk:
I'd suggest we make an images directory under "automation-good-practices/docs/"
i.e. automation-good-practices/docs/images

jscar-hawk:
Talked about it at the COP Meeting, and we are moving forward with Eric's solution for now.

jscar-hawk (approved):
Approved from the Automation-CoP Meeting.

Issue #21, opened 2021-07-21
RFE: please add instructions or Makefile to build PDF


pcfe:
in db2883884c26f53d20bcc6f4d1149b236735fc00 asciidoctor-pdf is mentioned.
It would be nice if the docs also showed users and/or potential contributors the command to use, just like they show how to generate HTML, or if a `Makefile` was added.

MR coming in a moment.

pcfe

ericzolf:
Not sure why this didn't happen automatically, but closed with #22

PR #22, opened 2021-07-21
introduce PDF generating and Makefile

pcfe:
addresses redhat-cop/automation-good-practices#21

This MR was created after thinking _oh, I'd like a PDF and I have a Makefile I regularly use anyway_.
Do tell if you would like creation of the HTML tree  added to the `Makefile` and your preferred target name if any.

`{gitdate}` and `{githash}` can be used in the AsciiDoc source

### DIFF
--- a/CONTRIBUTE.adoc
+++ b/CONTRIBUTE.adoc
@@ -98,7 +98,7 @@ Examples:: These are examples
 Even more examples...
 ====
 
-== Publishing
+== Publishing as HTML
 
 Use for now the following manual command:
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTE.adoc
```diff
@@ -98,7 +98,7 @@ Examples:: These are examples
 Even more examples...
 ====
 
-== Publishing
+== Publishing as HTML
```
ericzolf, line 101: It's really publishing, meaning to our website at https://redhat-cop.github.io/automation-good-practices/. Please rename more aptly to "Publish for the website".

ericzolf (changes requested):
I'm still not a big friend of makefiles but it can help the one or the other, so I'll aprove the PR once the title is fixed.

CONTRIBUTE.adoc
pcfe: done via 9be882a
feel free to drop 29581f4 I added that one as the link to the website was not immediately obvious to me when lookuing at the source repo.

ericzolf (approved):

jscar-hawk:
Automation-CoP has approved this PR on 2021-07-28

PR #29, opened 2021-09-24
Remove rule: Do not append `.j2` to template files

### LINKED ISSUE #16: remove rule that Jinja2 template files should not end in .j2
I think we should remove the following rule from the guidelines:

```
* When using the `template` module, refrain from appending `.j2` to the file name. This alters the syntax highlighting in most editors and will obscure the benefits of highlighting for the particular file type or filename.
  Anything under the `templates` directory of a role is assumed to be treated as a Jinja 2 template, so adding the `.j2` extension is redundant information that is not helpful.
```

**Reasoning**:

1. The given explanation that it removes syntax highlighting for the language used in the templated file in favour of Jinja2 syntax highlighting is actually a point _in favour_ of _adding_ the `.j2` file name suffix. 
When you're at the stage of writing a template file, you usually already know how the file should end up looking on the file system, so at that point the Jinja2 syntax is more important.
2. it is very easy to define in your editor settings to use, e.g., "html" syntax highlighting for all files ending in `.html.j2`. It is much harder to automatically enable Jinja2 syntax highlighting for _some_ files ending on `.html`.

mophahr:
Fixes #16

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -186,6 +186,16 @@ The use of `lineinfile` should include a comment with justification.
 +
 * Limit use of the `copy` module to copying remote files and to uploading binary blobs.
   For all other file pushes, use the `template` module. Even if there is nothing in the file that is being templated at the current moment, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` and then needs to be moved before it can be edited.
+* When using the `template` module, append `.j2` to the template file name. 
++
+[%collapsible]
+====
+Example:: If you want to use the `ansible.builtin.template` module to create a file called `example.conf` somewhere on the managed host, you should name the template for this file `templates/example.conf.j2`.
+Rationale:: When you are at the stage of writing a template file you usually already know how the file should end up looking on the file system, so at that point it is convenient to use Jinja2 syntax highlighting to make sure your templating syntax checks out.
+Should you need syntax highlighting for whatever language the target file should be in, it is very easy to define in your editor settings to use, e.g., HTML syntax highlighting for all files ending in `.html.j2`. 
+It is much less straightforward to automatically enable Jinja2 syntax highlighting for _some_ files ending on `.html`.
+====
++
 * Keep filenames and templates as close to the name on the destination system as possible.
 +
 [%collapsible]
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -186,8 +186,6 @@ The use of `lineinfile` should include a comment with justification.
 +
 * Limit use of the `copy` module to copying remote files and to uploading binary blobs.
   For all other file pushes, use the `template` module. Even if there is nothing in the file that is being templated at the current moment, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` and then needs to be moved before it can be edited.
-* When using the `template` module, refrain from appending `.j2` to the file name. This alters the syntax highlighting in most editors and will obscure the benefits of highlighting for the particular file type or filename.
-  Anything under the `templates` directory of a role is assumed to be treated as a Jinja 2 template, so adding the `.j2` extension is redundant information that is not helpful.
 * Keep filenames and templates as close to the name on the destination system as possible.
 +
 [%collapsible]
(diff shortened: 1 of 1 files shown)


ericzolf (approved):
I'm approving but I also asked the colleagues at https://chat.google.com/room/AAAATM-IFKw/rNBlR8fb1jE to give their feedback.

greg-hellings:
I disagree with this change. Addition of the `.j2` extension is redundant to the `templates` directory structure and introduces unnecessary cruft to the naming of files.

Also, hiding the syntax highlighting of the underlying file makes it very easy to introduce accidental errors to the structure of many document types such as XML or JSON files that are being templated.

greg-hellings (changes requested):
Please keep this rule

lotaris-xx:
Please remove this rule. It like saying we don't need ".c .h .py .whatever" because context is enough. .j2 is the simplest of indicators of author intent.

mophahr:
@greg-hellings @lotaris-xx thanks for your comments. I have implemented this change based on the feedback on #16 (the issue received a few 👍 s and no vetos, and we quickly discussed this in one of the CoP meetings)

ericzolf:
I have the feeling we shouldn't _remove_ the rule but _modify_ it and explain the pros and cons.

mophahr:
I have updated my pull request based on the discussion in and around the CoP meeting 2 weeks ago. Please have another look.

jscar-hawk:
I'd agree to add/keep the rule for keeping .j2 as part of the rules we would use with templating.

coding_style/README.adoc
```diff
@@ -186,8 +186,16 @@ The use of `lineinfile` should include a comment with justification.
 +
 * Limit use of the `copy` module to copying remote files and to uploading binary blobs.
   For all other file pushes, use the `template` module. Even if there is nothing in the file that is being templated at the current moment, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` and then needs to be moved before it can be edited.
-* When using the `template` module, refrain from appending `.j2` to the file name. This alters the syntax highlighting in most editors and will obscure the benefits of highlighting for the particular file type or filename.
-  Anything under the `templates` directory of a role is assumed to be treated as a Jinja 2 template, so adding the `.j2` extension is redundant information that is not helpful.
+* When using the `template` module, append `.j2` to the template file name. 
++
+[%collapsible]
+====
+Example:: If you want to use the `ansible.builtin.template` module to create a file called `example.conf` somewhere on the managed host, you should name the template for this file `templates/example.conf.j2`.
```
ericzolf, line 193:
```suggestion
Example:: If you want to use the `ansible.builtin.template` module to create a file called `example.conf` somewhere on the managed host, name the template for this file `templates/example.conf.j2`.
```
(avoid should!)

coding_style/README.adoc
```diff
@@ -186,8 +186,16 @@ The use of `lineinfile` should include a comment with justification.
...
+Rationale:: When you are at the stage of writing a template file you usually already know how the file should end up looking on the file system, so at that point it is convenient to use Jinja2 syntax highlighting to make sure your templating syntax checks out.
+Should you need syntax highlighting for whatever language the target file should be in, it is very easy to define in your editor settings to use, e.g., HTML syntax highlighting for all files ending in `.html.j2`.
```
ericzolf, line 195:
```suggestion
Should you need syntax highlighting for whatever language the target file should be in, it is very easy to define in your editor settings to use, e.g. HTML syntax highlighting for all files ending in `.html.j2`.
```

ericzolf (approved):
Two very small things. OK to merge.

coding_style/README.adoc
mophahr: Won't change the comma (see https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e) but the space will be fixed :)

ericzolf:
Fair enough :+1: https://theoatmeal.com/comics/ie

ericzolf:
Approved CoP meeting 2021-11-17

tonykay:
If the directory context is enough then by that logic we don't need .yml in `tasks`? The .j2 convention is widely used both inside the ansible community but also in the python developer community where .j2 are used widely eg for dynamic HTML etc. So it becomes artificial to say jinja files should end .j2 except when they are in a templates directory within a role. Finally, many developers editors from vim to vsc, are often customized to behave in different ways for different file types typically denoted via their extension (eg 2 spaces for .yml|yaml, 4 space for .py), the lack of .j2 breaks that common practice for no gain?

PR #33, opened 2021-10-23
Add recommendation to not use groups in roles

ericzolf:
or to at least make them a parameter of the role (aka defaults)

### DIFF
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -408,7 +408,7 @@ It is relatively common to use (inventory) group names in roles:
 * either to loop through the hosts in the group, generally in a cluster context
 * or to validate that a host is in a specific group
 +
-Use instead a variable (a list variable in the first case), or at least make thegroup's name a parameter of your role.
+Instead, store the host name(s) in a (list) variable, or at least make the group name a parameter of your role.
 You can always set the variable at group level to avoid repetitions.
 
 Rationale::
@@ -436,12 +436,12 @@ include::dont_use_groups/playbook.yml[]
 ----
 +
 The first approach is probably best to create a cluster configuration file listing all cluster's hosts.
-The other approaches are good to make sure each action applies only once, at the price of many skips.
+The other approaches are good to make sure each action is performed only once, but this comes at the price of many skips.
 The second one fails if the first host isn't reachable (which might be what you'd want anyway), and the last one has the best chance to be executed once and only once, even if some hosts aren't available.
 +
 TIP: the variable `cluster_group_name` could have a default group name value in your role, of course properly documented, for simple use cases.
 +
-Overall, it is best if you can avoid this kind of constructs as they are clumsy, but it depends on the use case if it's possible.
+Overall, it is best to avoid this kind of constructs if the use case permits, as they are clumsy.
 ====
 
 
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -413,7 +413,7 @@ You can always set the variable at group level to avoid repetitions.
 
 Rationale::
 Groups are a feature of the data in your inventory, meaning that you mingle data with code when you use those groups in your code.
-You should better rely on the inventory parsing process to provide your code with the variables it needs, not enforcing a specific structure of the inventory.
+Rely on the inventory-parsing process to provide your code with the variables it needs instead of enforcing a specific structure of the inventory.
 Not all inventory sources are flexible enough to provide exactly the expected group name.
 Even more importantly, in a cluster context for example, if the group name is fixed, you can't describe (and hence automate) more than one cluster in each inventory.
 On the other hand, variables can have any kind of value for each host, so that you can have as many clusters as you want.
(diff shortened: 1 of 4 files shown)


roles/README.adoc
```diff
@@ -399,6 +399,52 @@ described in the
...
++
+* either to loop through the hosts in the group, generally in a cluster context
+* or to validate that a host is in a specific group
++
+Use instead a variable (a list variable in the first case), or at least make thegroup's name a parameter of your role.
+You can always set the variable at group level to avoid repetitions.
+
+Rationale::
+Groups are a feature of the data in your inventory, meaning that you mingle data with code when you use those groups in your code.
+You should better rely on the inventory parsing process to provide your code with the variables it needs, not enforcing a specific structure of the inventory.
```
mophahr, line 416:
```suggestion
Rely on the inventory-parsing process to provide your code with the variables it needs instead of enforcing a specific structure of the inventory.
```

roles/README.adoc
```diff
@@ -399,6 +399,52 @@ described in the
...
+=== Don't use host group names or at least make them a parameter
+[%collapsible]
+====
+Explanations::
+It is relatively common to use (inventory) group names in roles:
++
+* either to loop through the hosts in the group, generally in a cluster context
+* or to validate that a host is in a specific group
++
+Use instead a variable (a list variable in the first case), or at least make thegroup's name a parameter of your role.
```
mophahr, line 411:
```suggestion
Instead, store the host name(s) in a (list) variable, or at least make the group name a parameter of your role.
```

roles/README.adoc
```diff
@@ -399,6 +399,52 @@ described in the
...
+You can always set the variable at group level to avoid repetitions.
+
+Rationale::
+Groups are a feature of the data in your inventory, meaning that you mingle data with code when you use those groups in your code.
+You should better rely on the inventory parsing process to provide your code with the variables it needs, not enforcing a specific structure of the inventory.
+Not all inventory sources are flexible enough to provide exactly the expected group name.
+Even more importantly, in a cluster context for example, if the group name is fixed, you can't describe (and hence automate) more than one cluster in each inventory.
```
mophahr, line 418:
```suggestion
Even more importantly, in a cluster context for example, if the group name is fixed, you can't describe (and hence automate) more than one cluster in each inventory becasue you cannot define multiple groups with the same name without having multiple inventories.
```

roles/README.adoc
```diff
@@ -399,6 +399,52 @@ described in the
...
+We can then use one of the following three approaches in our role (here as playbook, again for sake of simplicity):
++
+.A playbook showing how to loop through a group
+[source,yaml]
+----
+include::dont_use_groups/playbook.yml[]
+----
++
+The first approach is probably best to create a cluster configuration file listing all cluster's hosts.
+The other approaches are good to make sure each action applies only once, at the price of many skips.
```
mophahr, line 439:
```suggestion
The other approaches are good to make sure each action is performed only once, but this comes at the price of many skips.
```

roles/README.adoc
```diff
@@ -399,6 +399,52 @@ described in the
...
+The second one fails if the first host isn't reachable (which might be what you'd want anyway), and the last one has the best chance to be executed once and only once, even if some hosts aren't available.
++
+TIP: the variable `cluster_group_name` could have a default group name value in your role, of course properly documented, for simple use cases.
++
+Overall, it is best if you can avoid this kind of constructs as they are clumsy, but it depends on the use case if it's possible.
```
mophahr, line 444:
```suggestion
Overall, it is best to avoid this kind of constructs if the use case permits, as they are clumsy.
```

roles/dont_use_groups/playbook.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: show how to loop through a set of groups
+  hosts: cluster_group_?
+  gather_facts: false
+  become: false
+
+  tasks:
+  - name: the loop happens for each host, might be too much
```
mophahr, line 8: We have a rule to generally indent list elements (see https://github.com/redhat-cop/automation-good-practices/blob/mophahr--remove_rule_for_j2/coding_style/README.adoc#yaml-and-jinja2-syntax):

```suggestion
    - name: the loop happens for each host, might be too much
```
…and so on

mophahr (changes requested):
I have left some mostly cosmetic suggestions and corrections, but I think that the indentation needs to be consistent with our own recommendation in this guide.

jscar-hawk:
In the Automation-CoP Meeting this was Agreed upon, but we will check spelling and indentations before approving.

roles/README.adoc
ericzolf: That's IMHO an obvious statement (and there is a typo).

roles/dont_use_groups/playbook.yml
ericzolf: You're right. Done on the command line.

ericzolf:
All changes from @mophahr addressed. Thanks for the feedback.

mophahr (approved):
👍

ericzolf:
Approved CoP meeting 2021-11-17

Issue #32, opened 2021-10-12
Think about naming guideline for "generic" variables, used across multiple roles


ericzolf:
generic variables within or outside of a collection...

See https://chat.google.com/room/AAAAmYHFeFs/4sQz86ZPbEY for the (internal) discussion about the topic

PR #43, opened 2022-03-08
Started the collections readme

bontreger:


jscar-hawk:
Going to review the docs linked and will look at it at the next meeting

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
...
+====
+Explanations:: Collections should be comprised of roles collected either at the type or landscape level. See https://github.com/redhat-cop/automation-good-practices/blob/main/structures/README.adoc[the structures definition]
+
+Rationale:: Gathering and publishing collections, rather than individual roles, allows for easier distribution and particularly becomes more important when we discuss Execution Environments.
+
+====
+=== Importing Collections
+[%collapsible]
+====
+Explanations:: Collections should always be imported using requirements files within the role
```
ericzolf, line 16: After having read the linked article, I'm still not sure I understand which part of it you want to highlight, and from what I've read I'm not sure I agree with the statement.
I see two main statements in the blog article:
1. refer to collections dependencies (and roles) by name _and_ version
2. make sure to install roles and collections relative to your playbook/project/Git repo and _not_ in a central place

My issue and it's also due to a lack of experience so far, is how well those recommendations would work in the context of execution environments/controller/navigator:
1. how does controller/navigator react to collections not being installed where expected (or would the location be anyway overwritten)
2. how to handle conflicting versions between EE and requirements.yml?
3. an EE comes with a set version of some collections, so would your requirements.yml, you might need to change the version in two places
4. does navigator/controller recognize at all that a collection requested in requirements.yml is already available in the EE?

So, in short, we need to discuss all these aspects before we can take a decision.

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
 = Collections good practices
 
-NOTE: Work in Progress...
+Note: Unreviewed work. Please contribute to the discussion in the Automation Red Hat COP
+
+=== Collection Structure
```
ericzolf, line 5: The rule name should be a one-liner summarizing the rule, so that people can overlook at once the recommendations and only need to uncollapse the ones they need details for.

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
...
+====
+=== Importing Collections
+[%collapsible]
+====
+Explanations:: Collections should always be imported using requirements files within the role
+
+Rationale:: Prior art, https://www.jeffgeerling.com/blog/2020/ansible-best-practices-using-project-local-collections-and-roles
+====
+
+=== Collection Variables
```
ericzolf, line 21:
```suggestion
=== Create implicit collection variables and reference them in your roles' defaults variables
```
As explained, each rule title should summarize the recommendation.

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
...
+[%collapsible]
+====
+Explanations:: Often, variables will want to be defined on a collection level, but this can cause issues with roles being able to be reused.  By creating role specific and collection wide variables, this can be made clear and roles can remain reusable.
```
ericzolf, line 24:
```suggestion
Explanations:: Often, variables will want to be defined on a collection level, but this can cause issues with roles being able to be reused.
By defining collection wide variables and referencing them in roles' defaults variables, this can be made clear and roles can remain reusable.
Collection variables are nowhere defined explicitly and are to be documented in the collection's documentation.
```
We follow https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
...
+
+Rationale:: Variables that are shared across collections can cause collisions when roles are reused outside of the original collection.
```
ericzolf, line 26: I would add few more rationales:

1. role variables continue to be named according to our rules
2. it remains possible to overwrite the collection's variable value specifically for a role while globally define a value for the other collections
3. each role has its own set of default values for the variable

collections/README.adoc
```diff
@@ -1,3 +1,51 @@
...
+.Beta defaults/main.yml
+[source,yaml]
+----
+# specific role variables
+beta_job_name: 'some other text'
+# collection wide variables
+beta_controller_username: "{{ mycollection_controller_username }}"
+beta_no_log: "{{ mycollection_no_log | default('false') }}"
+----
+====
```
ericzolf, line 50: I would add a note admonition along the lines that there is no default for `controller_username` and it needs to be defined in one's inventory, whereas `no_log` has working default values.

ericzolf (changes requested):
It looks overall good and meaningful but there is still quite a lot to discuss (which proves that it's meaningful :smile:).
For the future, better create one PR for each new recommendation, it'll make reviewing/merging easier.

ericzolf:
Closes #32

ericzolf:
@bontreger could you find time to finalize this PR? It blocks #55

collections/README.adoc
bontreger: Agreed, removing for now

collections/README.adoc
bontreger: added this note

bontreger:
Ready for further review now.

ericzolf (approved):

ericzolf:
Thanks for your effort, looks good to me, I assume that we'll merge tomorrow.

mophahr:
merged in the CoP meeting

PR #44, opened 2022-03-09
Revert "Fix dead link"

ericzolf:
Reverts redhat-cop/automation-good-practices#40

Sorry, I have to revert: the link might be dead but you made an external reference from an internal one. This means that if you render the files locally (as HTML or even as PDF), this link would point to GitHub instead of pointing within your rendered content.


djdanielsson (approved):
it sounds like another solution needs to be found, but I will approve just to get this new issue fixed

PR #55, opened 2022-03-23
Mschoenw collection role name prefix

mophahr:
This is a follow up to the discussion we had in a recent CoP meeting.

The PR contains a rule to only use variable and role names in a collection that at least contain the collection name (or a short version of it)

### DIFF
--- a/collections/README.adoc
+++ b/collections/README.adoc
@@ -1,3 +1,53 @@
 = Collections good practices
 
 NOTE: Work in Progress...
+
+== Use "unique-enough" prefixes for all role and variable names within a collection
+
+[%collapsible]
+====
+Explanations:: Every role name in a collection should begin with a prefix that is reasonably unique to the collection.
+If the collection name is short enough, you can of course use it directly.
+For variables, either use the role name (or a short version of it) _and_ the collection name (or a short version of it)  as a prefix, or use the _collection name_ (or a short version of it)  only in case the variable are used in more than one of the collection's roles.
+
+Rationale:: This avoids confusion when using more than one collection, and it makes it easier to idetify where a role/variable comes from in projects where modules are called without the full collection name.
+
+Examples::
++
+.Use names like this
+[source,yaml]
+----
+  roles:
+    # example using role name and collection name
+    - role: redhat_cop.aap_utils.aap_utils_setup_prepare
+      vars:
+        - __aap_utils_setup_prepare_variable1
+        - aap_utils_setup_prepare_variable2
+        - __aap_utils_variable3
+        - aap_utils_variable4
+    # example using shorter versions
+    - role: redhat_cop.aap_utils.aap_setup_install
+      vars:
+        - __aap_setup_install_variable1
+        - aap_setup_install_variable2
+        - __aap_utils_variable3
+        - aap_utils_variable4
+        - __aap_install_variable5
+        - aap_install_variable4
+----
++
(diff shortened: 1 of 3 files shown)


ericzolf:
Was it on purpose that you mixed Makefile stuff and documentation?

mophahr:
> Was it on purpose that you mixed Makefile stuff and documentation?

nope. fixed.

ericzolf (changes requested):
We have a dependency sadly on #43 which isn't yet finished/merged.

mophahr:
> We have a dependency sadly on #43 which isn't yet finished/merged.

This shoiuld be ready now. @ericzolf please have another look

ericzolf (approved):

collections/README.adoc
```diff
@@ -1,17 +1,64 @@
...
 ====
 Explanations:: Collections should be comprised of roles collected either at the type or landscape level. See <<structures/README.adoc#_define_which_structure_to_use_for_which_purpose,The Structures Definition>>
 
 Rationale:: Gathering and publishing collections, rather than individual roles, allows for easier distribution and particularly becomes more important when we discuss Execution Environments.
+====
 
+== Use "unique-enough" prefixes for all role and variable names within a collection
+[%collapsible]
 ====
+Explanations:: Every role name in a collection should begin with a prefix that is reasonably similar to the collection name and unique.
```
sean-m-sullivan, line 15: Should be fixed to:
```
Explanations:: Every role and module name in a collection should begin with a name that is unique to the collection and descrbies what it does.
Examples:
    - redhat_cop.controller_configuration.labels
    - ansible.controller.job_template
    - redhat.rhel_system_roles.network


For Every variable, either use the collection name (or a short version of it) _and_ the role name (or a short version of it) as a prefix.
Example: __application_item.organization
Use the _collection name_ (or a short version of it)  only in case the variable are used in more than one of the collection's roles (see the rule below for details).
```

ericzolf: I think we're slightly making circles here. Let me try to summarize

- on variables, we're aligned I think and we've already merged something around [collection variables](https://github.com/redhat-cop/automation-good-practices/tree/main/collections#create-implicit-collection-variables-and-reference-them-in-your-roles-defaults-variables)
- we have already have generic rules for naming "things": https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#naming-things - I think we agree that they also apply to roles/modules within a collection, so we don't need to be specific
- so we only disagree on the collection related prefix for roles and modules:
    - both practices are used in the wild (see Azure with prefix, no prefix is probably more common)
    - I acknowledge the fact that a prefix is difficult to handle where upstream and downstream collections have different names (e.g. awx vs. controller vs. aap, or foreman vs. satellite)
    - I personally still find that a name like `host` (controller or Satellite?) is too generic and can be confusing out of context e.g. in an exception message, but I have to admit that it's a 
    - we already have a statement around using fully qualified role names (only for roles, in documentation only, and unrelated to collections, we probably need to extend the statement but other topic)
    - also against the prefix speaks the repetition within the FQCN

OK, big sigh but looking at the pros and cons, I have to admit that the prefix practice looks not so good as I thought it would. So let's scrap this PR altogether?

mophahr: Agreed. we can scap it.

mophahr:
closed during CoP meeting

PR #41, opened 2022-03-08
Rebuild roles document with new draft

bontreger:
Updated document to follow contribution standards and clarify the role practice recommendations

PR #61, opened 2022-04-06
Recreating changes from PR 41 for new PR

bontreger:
Recreating the changes from PR #41 , and removing previous PR

### DIFF
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -6,21 +6,26 @@ NOTE: this section has been rewritten, using https://github.com/oasis-roles/meta
 
 === Basic design
 [%collapsible]
+====
 Explanations:: Design roles focused on the functionality provided, not the software implementation.
 
 Rationale::
 Try to design roles focused on the functionality, not on the software implementation behind it.
 This will help abstracting differences between different providers, and help the user to focus on the functionality, not on technical details.
 
-Examples:: For an example, designing a role to implement an NTP configuration on a server would be a role.  The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
-However, when the underlying implementations become too divergent, for example implementing an email server with postfix or sendmail, then separate roles are encouraged.
+Examples:: 
+For an example, designing a role to implement an NTP configuration on a server would be a role.
+The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
+However, when the underlying implementations become too divergent, for example implementing an email server with postfix or sendmail, then 
+separate roles are encouraged.
 
 ====
 
 === Role Structure
 [%collapsible]
 ====
-Explanations:: New roles should be initiated in line, with the skeleton directory, which has standard boilerplate code for a Galaxy-compatible Ansible role and some enforcement around these standards
+Explanations:: New roles should be initiated in line, with the skeleton directory, which has standard boilerplate code for a Galaxy-compatible
+Ansible role and some enforcement around these standards
 
 Rationale:: A consistent file tree structure will help drive consistency and reusability across the entire environment.
 ====
@@ -28,9 +33,11 @@ Rationale:: A consistent file tree structure will help drive consistency and reu
 === Role Distribution
 [%collapsible]
 ====
-Explanations:: Use https://semver.org/[semantic versioning] for Git release tags.  Use 0.y.z before the role is declared stable (interface-wise).  
+Explanations:: Use https://semver.org/[semantic versioning] for Git release tags.  
+Use 0.y.z before the role is declared stable (interface-wise).  
 
-Rationale:: There are some https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible Galaxy and Automation Hub.  The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
+Rationale:: There are some https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible Galaxy and Automation Hub.
+The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
 
 === Naming parameters
 [%collapsible]
@@ -77,9 +84,15 @@ Exporting `$ROLENAME_provider_os_default` allows the users to set `$ROLENAME_pro
 === Distributions and Versions
 [%collapsible]
 ====
-Explanations:: Avoid testing for distribution and version in tasks. Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version. 
+Explanations:: Avoid testing for distribution and version in tasks.
+Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version. 
 
-Rationale:: This way it is easy to add support to a new distribution by simply dropping a new file in to "vars/", see below <<supporting-multiple-distributions-and-versions,Supporting multiple distributions and versions>>. See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding large lists or 'magic values' directly into the playbook." Since distribution-specific values are kind of "magic values", it applies to them. The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed. Consider making paths to templates internal variables if you need different templates for different distributions.
+Rationale:: 
+This way it is easy to add support to a new distribution by simply dropping a new file in to "vars/", see below <<supporting-multiple-distributions-and-versions,Supporting multiple distributions and versions>>.
+See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding large lists or 'magic values' directly into the playbook." 
+Since distribution-specific values are kind of "magic values", it applies to them.
+The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
+Consider making paths to templates internal variables if you need different templates for different distributions.
 ====
 
 === Check Mode
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -11,6 +11,8 @@ Explanations:: Design roles focused on the functionality provided, not the softw
 Rationale:: Try to design roles focused on the functionality, not on the software implementation behind it. This will help abstracting differences between different providers, and help the user to focus on the functionality, not on technical details.
 
 Examples:: For an example, designing a role to implement an NTP configuration on a server would be a role.  The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
+However, when the underlying implementations become too divergent, for example implementing an email server with postfix or sendmail, then separate roles are encouraged.
+
 ====
 
 === Role Structure
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -1,63 +1,35 @@
 = Roles Good Practices for Ansible
 
-NOTE: this section has been imported "as-is" from the https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and still requires re-formatting to fit the overall structure.
+NOTE: this section has been rewritten, using https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and https://docs.google.com/document/d/1jZUsman8IDSvDt-s6loJ3Y4uZsyse3CB-xzM2DCYaHQ/edit[Ansible Good Practice Recommendations] as a starting place. If you have anything to add or review, please comment.
 
-== Background
-[%collapsible]
-====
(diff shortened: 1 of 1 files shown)


roles/README.adoc
```diff
@@ -1,63 +1,35 @@
...
-* Use https://semver.org/[semantic versioning] for Git release tags.
-  Use 0.y.z before the role is declared stable (interface-wise).
-  Although it has not been a problem so far for linux system roles, since they use strict X.Y.Z versioning, you should be aware that there are some
-https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible
-Galaxy and Automation Hub.
-  The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
-====
-
-== Interface design considerations
+Explanations:: New roles should be initiated in line, with the skeleton directory, which has standard boilerplate
```
mophahr, line 19: quick comment: we generally do one line per sentence in this document.

roles/README.adoc
```diff
@@ -1,63 +1,35 @@
...
-
-Designing the interface in the latter case is difficult because it has to be sufficiently abstract to cover different providers.
-We, for example, do not provide an email role in the Linux System Roles project, only a postfix role, because the underlying implementations (sendmail, postfix) were deemed to be too divergent.
-Generally, an abstract interface should be something that should be always aimed for though, especially if there are multiple providers in use already, and in
-particular when the default provider is changing or is known to be likely to change in the next major releases.
+Explanations:: Design roles focused on the functionality provided, not the software implementation.
+
+Rationale:: Try to design roles focused on the functionality, not on the software implementation behind it. This will help abstracting differences between different providers, and help the user to focus on the functionality, not on technical details.
+
+Examples:: For an example, designing a role to implement an NTP configuration on a server would be a role.  The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
```
ericzolf, line 13: I would add back the example about mail server, explaining that there are circumstances where such an approach doesn't make sense because too different.

ericzolf (changes requested):
Mainly formatting should be improved (one sentence per line), else looks good to me, with one little comment (to be discussed).

bontreger (commented):
Addressed comments

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
 = Roles Good Practices for Ansible
 
-NOTE: this section has been imported "as-is" from the https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and still requires re-formatting to fit the overall structure.
+NOTE: this section has been rewritten, using https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and https://docs.google.com/document/d/1jZUsman8IDSvDt-s6loJ3Y4uZsyse3CB-xzM2DCYaHQ/edit[Ansible Good Practice Recommendations] as a starting place. If you have anything to add or review, please comment.
```
mophahr, line 3: The google doc isn't public. Does ist still make sense to link to it?

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
-* Use https://semver.org/[semantic versioning] for Git release tags.
-  Use 0.y.z before the role is declared stable (interface-wise).
-  Although it has not been a problem so far for linux system roles, since they use strict X.Y.Z versioning, you should be aware that there are some
-https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible
-Galaxy and Automation Hub.
-  The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
-====
-
-== Interface design considerations
+Explanations:: New roles should be initiated in line, with the skeleton directory, which has standard boilerplate code for a Galaxy-compatible Ansible role and some enforcement around these standards
```
mophahr, line 21: Don't we have a rule about deleting empty files/directories somewhere? I'm not 100% sure r/n @ericzolf do you happen to know?

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
 
-NOTE: this section has been imported "as-is" from the https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and still requires re-formatting to fit the overall structure.
+NOTE: this section has been rewritten, using https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and https://docs.google.com/document/d/1jZUsman8IDSvDt-s6loJ3Y4uZsyse3CB-xzM2DCYaHQ/edit[Ansible Good Practice Recommendations] as a starting place. If you have anything to add or review, please comment.
 
-== Background
+== Role design considerations
+
+=== Basic design
 [%collapsible]
-====
```
mophahr, line 7: This line is still needed

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
-** Users who want to switch to a different provider in the same version of the OS (like switching from ntpd to chrony to RHEL 7) and keep the same playbook.
-
-Designing the interface in the latter case is difficult because it has to be sufficiently abstract to cover different providers.
-We, for example, do not provide an email role in the Linux System Roles project, only a postfix role, because the underlying implementations (sendmail, postfix) were deemed to be too divergent.
-Generally, an abstract interface should be something that should be always aimed for though, especially if there are multiple providers in use already, and in
-particular when the default provider is changing or is known to be likely to change in the next major releases.
+Explanations:: Design roles focused on the functionality provided, not the software implementation.
+
+Rationale:: Try to design roles focused on the functionality, not on the software implementation behind it. This will help abstracting differences between different providers, and help the user to focus on the functionality, not on technical details.
```
mophahr, line 11:
```suggestion
Rationale:: Try to design roles focused on the functionality, not on the software implementation behind it.
This will help abstracting differences between different providers, and help the user to focus on the functionality, not on technical details.
```

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
+
+Examples:: For an example, designing a role to implement an NTP configuration on a server would be a role.  The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
```
mophahr, line 13:
```suggestion
Examples:: For an example, designing a role to implement an NTP configuration on a server would be a role.
The role internally would have the logic to decide whether to use ntpd, chronyd, and the ntp site configurations.
```

roles/README.adoc
mophahr, line 3:
```suggestion
NOTE: this section has been rewritten, using https://github.com/oasis-roles/meta_standards[OASIS metastandards repository] and https://docs.google.com/document/d/1jZUsman8IDSvDt-s6loJ3Y4uZsyse3CB-xzM2DCYaHQ/edit[Ansible Good Practice Recommendations] as a starting place.
If you have anything to add or review, please comment.
```

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
+====
 
-=== Basic design
+=== Role Distribution
 [%collapsible]
 ====
-Try to design the interface focused on the functionality, not on the software implementation behind it.
-This will help abstracting differences between different providers (see above), and help the user to focus on the functionality, not on technical details.
-====
+Explanations:: Use https://semver.org/[semantic versioning] for Git release tags.  Use 0.y.z before the role is declared stable (interface-wise).
```
mophahr, line 29:
```suggestion
Explanations:: Use https://semver.org/[semantic versioning] for Git release tags.
Use 0.y.z before the role is declared stable (interface-wise).  
```

roles/README.adoc
```diff
@@ -1,63 +1,36 @@
...
-=== Basic design
+=== Role Distribution
 [%collapsible]
 ====
-Try to design the interface focused on the functionality, not on the software implementation behind it.
-This will help abstracting differences between different providers (see above), and help the user to focus on the functionality, not on technical details.
-====
+Explanations:: Use https://semver.org/[semantic versioning] for Git release tags.  Use 0.y.z before the role is declared stable (interface-wise).  
+
+Rationale:: There are some https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible Galaxy and Automation Hub.  The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
```
mophahr, line 31:
```suggestion
Rationale:: There are some https://github.com/ansible/ansible/issues/67512[restrictions] for Ansible Galaxy and Automation Hub.
The versioning must be in strict X.Y.Z[ab][W] format, where X, Y, and Z are integers.
```

roles/README.adoc
```diff
@@ -99,23 +72,15 @@ Moreover, after a major upgrade of their systems, it may force the users to chan
...
-Avoid testing for distribution and version in tasks.
-Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version.
-This way it is easy to add support to a new distribution by
-simply dropping a new file in to "vars/", see below
-<<supporting-multiple-distributions-and-versions,Supporting multiple distributions and versions>>.
-See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding large lists or 'magic values' directly into the playbook."
-Since distribution-specific values are kind of "magic values", it applies to them.
-The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
-Consider making paths to templates internal variables if you need different templates for different distributions.
+Explanations:: Avoid testing for distribution and version in tasks. Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version.
```
mophahr, line 78:
```suggestion
Explanations:: Avoid testing for distribution and version in tasks.
Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version. 
```

roles/README.adoc
```diff
@@ -99,23 +72,15 @@ Moreover, after a major upgrade of their systems, it may force the users to chan
...
-Consider making paths to templates internal variables if you need different templates for different distributions.
+Explanations:: Avoid testing for distribution and version in tasks. Rather add a variable file to "vars/" for each supported distribution and version with the variables that need to change according to the distribution and version. 
+
+Rationale:: This way it is easy to add support to a new distribution by simply dropping a new file in to "vars/", see below <<supporting-multiple-distributions-and-versions,Supporting multiple distributions and versions>>. See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding large lists or 'magic values' directly into the playbook." Since distribution-specific values are kind of "magic values", it applies to them. The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed. Consider making paths to templates internal variables if you need different templates for different distributions.
```
mophahr, line 80:
```suggestion
Rationale:: This way it is easy to add support to a new distribution by simply dropping a new file in to "vars/", see below <<supporting-multiple-distributions-and-versions,Supporting multiple distributions and versions>>.
See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding large lists or 'magic values' directly into the playbook."
Since distribution-specific values are kind of "magic values", it applies to them.
The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
Consider making paths to templates internal variables if you need different templates for different distributions.
```

roles/README.adoc
```diff
@@ -140,15 +105,21 @@ Consider making paths to templates internal variables if you need different temp
   If this is to be supported, see https://github.com/linux-system-roles/timesync/issues/27#issuecomment-472466223[hhaniel's proposal], which seems to properly guard even against such cases.
 ====
 
+=== Idempotency
+[%collapsible]
+====
+Explanations:: Reporting changes properly is related to the other requirement: *idempotency*. Roles should not perform changes when applied a second time to the same system with the same parameters, and it should not report that changes have been done if they have not been done. Due to this, using `command:` is problematic, as it always reports changes. Therefore, override the result by using `changed_when:`
```
mophahr, line 111:
```suggestion
Explanations:: Reporting changes properly is related to the other requirement: *idempotency*.
Roles should not perform changes when applied a second time to the same system with the same parameters, and it should not report that changes have been done if they have not been done.
Due to this, using `command:` is problematic, as it always reports changes.
Therefore, override the result by using `changed_when:`
```

mophahr (dismissed):
I have added a few suggestions (mostly about formatting rules)

roles/README.adoc
richm: > The google doc isn't public. Does ist still make sense to link to it?

No.  This should be omitted.  And - that google doc should be updated to point to this public doc instead.

ericzolf:
@bontreger the changes required by @mophahr look good to me, and I'd be happy to approve once applied.

ericzolf (approved):

ericzolf:
Even if @mophahr won't be in tomorrow's session, I feel like we could merge it, but to be agreed with the team.

ericzolf:
Approved in CoP meeting 2022-06-29

PR #62, opened 2022-04-11
Use set_vars.yml for OS/platform specific vars

richm:
Use `tasks/set_vars.yml` to set OS/platform specific vars.  Using
a separate file allows role integration tests to access the
OS/platform specific vars.
Handle the case where a user specifies `gather_facts: false` in
the playbook - have the role gather only the facts it requires.

### DIFF
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -141,16 +141,60 @@ Consider making paths to templates internal variables if you need different temp
 ====
 
 === Supporting multiple distributions and versions
+Use Cases::
+* The role developer needs to be able to set role variables to different values
+depending on the OS platform and version.  For example, if the name of a service
+is different between EL8 and EL9, or a config file location is different.
+* The role developer needs to handle the case where the user specifies
+`gather_facts: false` in the playbook.
+* The role developer needs to access the platform specific vars in role
+integration tests without making a copy.
+
+NOTE: The recommended solution below requires at least some `ansible_facts` to
+be defined, and so relies on gathering some facts.  If you just want to ensure
+the user always uses `gather_facts: true`, and do not want to handle this in the
+role, then the role documentation should state that `gather_facts: true` or
+`setup:` is required in order to use the role, and the role should use `fail:`
+with a descriptive error message if the necessary facts are not defined.
+
+If it is desirable to use roles that require facts, but fact gathering is
+expensive, consider using a cache plugin
+https://docs.ansible.com/ansible/latest/collections/index_cache.html[List of Cache Plugins], and also consider running a periodic job on the controller to
+refresh the cache.
 
 ==== Platform specific variables
 [%collapsible]
 ====
 You normally use `vars/main.yml` (automatically included) to set variables
 used by your role.
-If some variables need to be parameterized according to distribution and version (name of packages, configuration file paths, names of services), use this in the beginning of your `tasks/main.yml`:
+If some variables need to be parameterized according to distribution and version (name of packages, configuration file paths, names of services), do the following:
+
+* Add the following to `vars/main.yml` (create the file if it does not exist):
 
 [source,yaml]
 ----
+# ansible_facts required by the role
+__rolename_required_facts:
+  - distribution
+  - distribution_major_version
+  - distribution_version
+  - os_family
+----
+
+You may need to add more facts, depending on what Ansible facts your role
+requires.  Try to add only the facts required by the role.  If you add more
+facts, you may need to change `gather_subset: min` - see below.
+
+* Create a file called `tasks/set_vars.yml` with the following contents:
+
+[source,yaml]
+----
+- name: Ensure ansible_facts used by role
+  setup:
+    gather_subset: min
+  when: not ansible_facts.keys() | list |
+    intersect(__rolename_required_facts) == __rolename_required_facts
+
 - name: Set platform/version specific variables
   include_vars: "{{ __rolename_vars_file }}"
   loop:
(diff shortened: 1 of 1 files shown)


richm:
See https://github.com/linux-system-roles/storage/pull/262 and https://github.com/linux-system-roles/storage/pull/260 for examples.

roles/README.adoc
```diff
@@ -163,6 +207,16 @@ If some variables need to be parameterized according to distribution and version
   when: __rolename_vars_file is file
 ----
 
+* Add this as the first task in `tasks/main.yml`:
+
```
branic, line 211: There should be a `+` on this line so the code block is lined up with the bullet.

branic (changes requested):
Other than the minor change to add the `+` in the locations indicated I think this looks good.

bontreger:
Are these three use cases independent? (if any one of these cases is true, then this is a recommendation) or together (if all 3 are true).

I feel like the note, roles that require facts to be gathered should state so explicitly, or fail if undefined is the real good practice here.

richm:
> Are these three use cases independent? (if any one of these cases is true, then this is a recommendation) or together (if all 3 are true).

The 3 cases are independent.  

Yeah - not sure how to do this - for example, if all you want to do is make the role work when `gather_facts: false`, this may or may not have anything to do with setting os/platform version specific variables.  However, the primary way that roles use facts is to handle os/platform version specific conditionals (such as setting variables) (at least in the case of linux-system-roles).  Any suggestions? 

> 
> I feel like the note, roles that require facts to be gathered should state so explicitly, or fail if undefined is the real good practice here.

Brian C. and I discussed this at length - his point was that roles should just `fail: msg: "You must use gather_facts: or setup: in order to use this role" when: distribution is not defined` - IMO that UX is not as good as the role gathering the facts it requires - and using cache plugins to cache fact to avoid gathering facts as much as possible will help in either case.

roles/README.adoc
```diff
@@ -141,16 +141,60 @@ Consider making paths to templates internal variables if you need different temp
...
 ==== Platform specific variables
 [%collapsible]
 ====
 You normally use `vars/main.yml` (automatically included) to set variables
 used by your role.
-If some variables need to be parameterized according to distribution and version (name of packages, configuration file paths, names of services), use this in the beginning of your `tasks/main.yml`:
+If some variables need to be parameterized according to distribution and version (name of packages, configuration file paths, names of services), do the following:
+
+* Add the following to `vars/main.yml` (create the file if it does not exist):
+
```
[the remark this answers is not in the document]
richm: > There should be a `+` on this line so the code block is lined up with the bullet.

Sorry - I'm totally unfamiliar with asciidoc - should it look like this?
```
+* Add the following to `vars/main.yml` (create the file if it does not exist):
```
?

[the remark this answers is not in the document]
richm: I got it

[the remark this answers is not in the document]
branic: Not quite. Line 173 should have the `+`. So lines 172-175 would look like this:

```
* Add the following to `vars/main.yml` (create the file if it does not exist):
+
[source,yaml]
----
```

roles/README.adoc
```diff
@@ -141,16 +141,60 @@ Consider making paths to templates internal variables if you need different temp
...
+[source,yaml]
+----
+# ansible_facts required by the role
+__rolename_required_facts:
+  - distribution
+  - distribution_major_version
+  - distribution_version
+  - os_family
+----
+
```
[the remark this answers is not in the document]
branic: @richm Should the paragraph beginning on line 184 still be part of (e.g. lined up with) the first bullet (it seems like it should be to me)? If so can you add a `+` on line 183.

branic (approved):

roles/README.adoc
```diff
@@ -141,16 +141,60 @@ Consider making paths to templates internal variables if you need different temp
 ====
 
 === Supporting multiple distributions and versions
+Use Cases::
+* The role developer needs to be able to set role variables to different values
```
mophahr, line 145: Content-wise, I have no objections r/n, but please have a look at the [contribution guide](https://github.com/redhat-cop/automation-good-practices/blob/main/CONTRIBUTE.adoc); 
it states "We follow the [Asciidoc recommended practices](https://asciidoctor.org/docs/asciidoc-recommended-practices/).", and that in turn states "Don’t wrap text at a fixed column width. Instead, put each sentence on its own line, a technique called sentence per line."

mophahr (changes requested):
plaese have a look at my formatting suggestion

roles/README.adoc
richm: done

mophahr (approved):

roles/README.adoc
```diff
@@ -203,6 +244,35 @@ If this is a problem, construct the file list as a list variable, and filter the
 ----
 
 Or define your `__rolename_vars_file_list` in your `vars/main.yml`.
+
+The task `Ensure ansible_facts used by role` handles the case where the user specifies `gather_facts: false` in the playbook.
+It gathers *only* the facts required by the role.
+The role developer may need to add additional facts to the list, and use a different `gather_subset`.
+See https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html#setup-module[Setup
+Module] for more information.
```
mophahr, line 252:
```suggestion
See https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html#setup-module[Setup Module] for more information.
```

richm: done

richm:
Can this be merged now?  Any other issues?

mophahr:
> Can this be merged now? Any other issues?

I'd suggest we do that in the next CoP meeting @richm

mophahr:
Approved to be merged in the ACoP meeting

PR #63, opened 2022-05-18
Add recommendations for plugin development

jillr:
Draft of plugin recommendations from ansible eng

### DIFF
--- a/plugins/README.adoc
+++ b/plugins/README.adoc
@@ -221,3 +221,23 @@ https://docs.ansible.com/ansible/latest/network/user_guide/network_resource_modu
 
 ====
 
+== Use clear error/info messages
+[%collapsible]
+====
+Explanations:: This will make it easier to troubleshoot failures if they occur
+
+Rationale:: Error messages that communicate specific details of the failure will aid in resolving the problem.
+Unclear error messages such as "Failed!" are unnecessarily obscure.
+
+Examples::
+[source,python]
+----
+    if checksum and checksum_src != checksum:
+        module.fail_json(
+            msg='Copied file does not match the expected checksum. Transfer failed.',
+            checksum=checksum_src,
+            expected_checksum=checksum
+        )
+----
+
+====
--- a/plugins/README.adoc
+++ b/plugins/README.adoc
@@ -12,23 +12,34 @@ NOTE: Work in Progress...
 == All plugin types should use argspec validation, inbound and outbound
 [%collapsible]
 ===
-Explanations::
+Explanations:: A plugin's `argument_spec` (or "argspec") defines the supported arguments for a module, as well as their type, defaults and more.
 
-Rationale::
+Rationale:: Validating the argspec at the time that it enters and exits the code ensures that these inputs and outputs conform to the plugin's documentation and the user's expectations.
 
 Examples::
+From the https://github.com/ansible-collections/ansible.utils/blob/2.6.0/plugins/filter/ipaddr.py[ipaddr] filter plugin.
+[source,python]
+----
+from ansible_collections.ansible.utils.plugins.module_utils.common.argspec_validate import (
+    AnsibleArgSpecValidator,
+)
 
+
+aav = AnsibleArgSpecValidator(
+    data=data, schema=DOCUMENTATION, name="ipaddr"
+)
+valid, errors, updated_data = aav.validate()
+if not valid:
+    raise AnsibleFilterError(errors)
+return ipaddr(**updated_data)
+----
 ====
 
 == All plugin types should have documentation
 [%collapsible]
 ====
 Explanations::
-
-Rationale::
-
-Examples::
-
+All plugins, regardless of type, should have documentation that describes the input parameters, outputs, and practical examples of how to use it.
 ====
 
 == A plugin’s argspec should not be incrementally built but instead, be defined as a single object
@@ -126,36 +137,39 @@ def test_split_pre_existing_dir_one_level_exists(directory, expected, mocker):
 == A complex plugin argspec should be derived from its documentation runtime if the plugin’s source is manually maintained.
 [%collapsible]
 ====
-Explanations:: A complex plugin argspec should be derived from its documentation runtime if the plugin’s source is manually maintained.  ( >30 keys or depth>2) (action plugins and control node execution modules)
+Explanations::
+A complex plugin argspec should be derived from its documentation runtime if the plugin’s source is manually maintained.  ( >30 keys or depth>2) (action plugins and control node execution modules)
 
-Rationale:: Hand-crafting large and complex argument specifications can be error-prone and difficult to maintain.  Programmatically generating the spec from DOCUMENTATION ensures consistency and accuracy. 
-
-Examples::
-
+Rationale::
+Hand-crafting large and complex argument specifications can be error-prone and difficult to maintain.  Programmatically generating the spec from DOCUMENTATION (already a structured data object) ensures consistency and accuracy. 
 
 ====
 
 == A plugin’s entry file should not contain more than 500 lines
 [%collapsible]
 ====
-Explanations:: A plugin’s entry file should not contain more than 500 lines
+Explanations::
+A plugin’s entry file should not contain more than 500 lines
 
 Rationale::
-
-Examples::
-
-
+Long and complex code files can be difficult to maintain.
+Wherever possible, reusable functions and classes, such as those for data validation or manipulation, should be moved to module_utility or plugin_utility files and imported into plugins.
+This keeps the plugin code easier to read and maintain.
+500 lines is a general guideline that should serve most cases.
 ====
 
 == Action plugins should be used when the code runs on the control node.
 [%collapsible]
 ====
 Explanations::
-
+Use an action plugin for code that needs to run on the controller, where the host would be `localhost`.
 Rationale::
+Action plugins let you integrate local processing and local data with module functionality.
+Action plugins act in conjunction with modules to execute the actions required by playbook tasks.
+Using an action plugin will cause Ansible to skip creating a temporary tarball of execution code and runs the code directly on the controller host.
 
 Examples::
-
+See the `ansible.snmp.get` https://github.com/ansible-network/ansible.snmp/blob/main/plugins/action/get.py[action plugin] and associated https://github.com/ansible-network/ansible.snmp/blob/main/plugins/modules/get.py[module].
 
 ====
 
@@ -175,11 +189,7 @@ Examples::
 [%collapsible]
 ====
 Explanations::
-
-Rationale::
-
-Examples::
-
+The https://github.com/ansible-community/ansible.plugin_builder[ansible.plugin_builder] is a tool which helps developers scaffold new plugins.
 
 ====
 
--- a/plugins/README.adoc
+++ b/plugins/README.adoc
@@ -42,18 +42,6 @@ Explanations::
 All plugins, regardless of type, should have documentation that describes the input parameters, outputs, and practical examples of how to use it.
 ====
 
-== A plugin’s argspec should not be incrementally built but instead, be defined as a single object
-[%collapsible]
-====
-Explanations::
-
-Rationale::
-
-Examples::
-
-
-====
-
 == Use sphinx (reST) formatted docstrings
 [%collapsible]
 ====
@@ -145,6 +133,27 @@ Hand-crafting large and complex argument specifications can be error-prone and d
 
 ====
 
+== Formatting of manually maintained plugin argspecs
+[%collapsible]
+====
+Explanations::
+Ensure a consistent approach to the way commplex argument_specs are formatted within a collection.
+
+Rationale::
+When hand-writing a complex argspec, the author may choose to build up to data structure from multiple dictionaries or vars.
+Other authors may choose to implement a complex, nested argspec as a single dictionary.
+Within a single collection, select one style and use it consistently.
+
+Examples::
+Single dictionary:
+https://github.com/ansible-collections/cisco.nxos/blob/3.0.0/plugins/module_utils/network/nxos/argspec/bgp_global/bgp_global.py
+
+Multiple dictionaries:
+https://github.com/ansible-collections/community.aws/blob/stable-3/plugins/modules/ec2_scaling_policy.py#L355-L370
+https://github.com/ansible-collections/amazon.cloud/blob/0.1.0/plugins/modules/backup_report_plan.py#L182-L234
+====
+
+
 == A plugin’s entry file should not contain more than 500 lines
 [%collapsible]
 ====
@@ -177,11 +186,11 @@ See the `ansible.snmp.get` https://github.com/ansible-network/ansible.snmp/blob/
 [%collapsible]
 ====
 Explanations::
+A plugin should be developed in accordance with Ansible best practices, without accounting for possible environment-specific variables.
 
 Rationale::
-
-Examples::
-
+The plugin author should prioritize development of the plugin's function.
+If an environment that the plugin could possibly run in (such as on a jump host) has some unique property this should be managed by the operator's Ansible configuration, separate from the plugin code.
 
 ====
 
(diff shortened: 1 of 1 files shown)


jillr:
@cidrblock A lot of word smithing will be needed to add Explanations, Rational, and Examples to the bullet list of recommendations we drafted.  I've taken a start here so we can revise in reviews.

plugins/README.adoc
```diff
@@ -7,3 +7,237 @@ NOTE: Work in Progress...
...
+====
+Explanations::
+
+Rationale::
+
+Examples::
+
+====
+
+== A plugin’s argspec should not be incrementally built but instead, be defined as a single object
```
jillr, line 34: @cidrblock I'm questioning if we need to keep this recommendation given the later one:
> A complex plugin argspec should be derived from its documentation runtime if the plugin’s source is manually maintained.

If a complex argspec is not being automatically generated, I would disagree with this recommendation (there are cloud examples where trying to hand-write complex nested argspecs without splitting pieces out into vars is a huge pain).  I think generating complex specs is the better advice between then though.

What do you think?

cidrblock: IMHO...

If the arspec is mantained by a human, I generally prefer:
1) That it be built runtime from the docstring, this ensures there is a single SOT for it (yeah, ansible-test does compare, but it still means double the updates)
2) If it can't be derived run time from the docstring (there are some limitations here and I've only seen it done to date in a limited fashion) I personally prefer a single data structure like this: https://github.com/ansible-collections/cisco.nxos/blob/main/plugins/module_utils/network/nxos/argspec/bgp_global/bgp_global.py rather than an incrementally built up data structure using a slew of `dict()` statements or even this incremental approach https://github.com/ansible-collections/amazon.cloud/blob/main/plugins/modules/backup_report_plan.py#L190

That said, both work, ( although I think I recall `dict()` being significantly slower than `{}`) and this is probably more of an individual dev preference.  I suspect some teams will decide on a pattern.

Maybe, "Coding styles for argspec definition should be consistent within a collection?"

jillr: With deeply nested options dictated by a 3rd party API we've seen folks have difficulty with properly formatting the dictionaries and it creates a lot of frustrating back and forth in PRs and seriously ugly, difficult to review arg specs (AWS Auto Scaling comes to mind).  It's certainly a huge argument in favor of generating specs, but we have the code we have today.  :)

"Consistent within a collection" works for me.

plugins/README.adoc
```diff
@@ -7,3 +7,247 @@ NOTE: Work in Progress...
...
+
+== All plugin types should use argspec validation, inbound and outbound
+[%collapsible]
+===
+Explanations:: A plugin's `argument_spec` (or "argspec") defines the supported arguments for a module, as well as their type, defaults and more.
+
+Rationale:: Validating the argspec at the time that it enters and exits the code ensures that these inputs and outputs conform to the plugin's documentation and the user's expectations.
+
+Examples::
+From the https://github.com/ansible-collections/ansible.utils/blob/2.6.0/plugins/filter/ipaddr.py[ipaddr] filter plugin.
```
NilashishC, line 20: I don't know if we have an example ready, but maybe we can also validate the return payload by comparing it with the `RETURN` defined in the plugin file. That way we can guarantee that the plugin returns _exactly_ what it says it does.

plugins/README.adoc
```diff
@@ -7,3 +7,247 @@ NOTE: Work in Progress...
...
+Examples::
+
+
+====
+
+== Use sphinx (reST) formatted docstrings
+[%collapsible]
+====
+Explanations::
+All classes, private and public functions should have a sphinx (reST) formatted docstring including all parameters, yields, raises, or returns.
```
NilashishC, line 61: +100

plugins/README.adoc
```diff
@@ -7,3 +7,247 @@ NOTE: Work in Progress...
...
+
+== Action plugins should be used when the code runs on the control node.
+[%collapsible]
+====
+Explanations::
+Use an action plugin for code that needs to run on the controller, where the host would be `localhost`.
+Rationale::
+Action plugins let you integrate local processing and local data with module functionality.
+Action plugins act in conjunction with modules to execute the actions required by playbook tasks.
+Using an action plugin will cause Ansible to skip creating a temporary tarball of execution code and runs the code directly on the controller host.
```
NilashishC, line 169: Should this be enforced for platforms where the ability to skip this is already present (aka `import_modules`)? What other benefits does implementing as an action plugin have in such cases?

plugins/README.adoc
ganeshrn: @NilashishC +1 for validating the return payload for the module. Maybe this can be in the ansible-core argspec validation.

plugins/README.adoc
jillr: Are there changes we should make to this section?  I'm happy to update the verbiage or add/remove things as needed - I'm less familiar with the networking use-case discussed so I'm not sure if there's changes needed.

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+
+Examples::
+See the `ansible.snmp.get` https://github.com/ansible-network/ansible.snmp/blob/main/plugins/action/get.py[action plugin] and associated https://github.com/ansible-network/ansible.snmp/blob/main/plugins/modules/get.py[module].
+
+====
+
+== Design decisions accounting for “jump hosts” should be avoided.
```
jillr, line 185: I was struggling with writers block on this one.  If anyone would like to see different verbiage here I welcome suggestions.

ganeshrn (approved):

justjais (approved):

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+Single dictionary:
+https://github.com/ansible-collections/cisco.nxos/blob/3.0.0/plugins/module_utils/network/nxos/argspec/bgp_global/bgp_global.py
+
+Multiple dictionaries:
+https://github.com/ansible-collections/community.aws/blob/stable-3/plugins/modules/ec2_scaling_policy.py#L355-L370
+https://github.com/ansible-collections/amazon.cloud/blob/0.1.0/plugins/modules/backup_report_plan.py#L182-L234
+====
+
+
+== A plugin’s entry file should not contain more than 500 lines
```
ericzolf, line 157: I would keep the concrete number out of the title and explanation, and only keep it in the rationale. You could write that it needs to have a "handy size", and then you can avoid "should" :smirk:

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+[%collapsible]
+====
+Explanations::
+A plugin’s entry file should not contain more than 500 lines
+
+Rationale::
+Long and complex code files can be difficult to maintain.
+Wherever possible, reusable functions and classes, such as those for data validation or manipulation, should be moved to module_utility or plugin_utility files and imported into plugins.
```
ericzolf, line 165: Difference between module_utility and plugin_utility?

plugins/README.adoc
ericzolf: I'm not sure to understand what you're trying to express. You seem to be hinting at a specific bad decision that you'd like to avoid in the future, and I'm probably missing this context. A lot of modules use environment variables, so I don't know why it should be a bad thing overall.

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+
+== Plugins should be initially developed using the ansible plugin builder
+[%collapsible]
+====
+Explanations::
+The https://github.com/ansible-community/ansible.plugin_builder[ansible.plugin_builder] is a tool which helps developers scaffold new plugins.
+
+====
+
+== Pylint and type exceptions should have a specific exclusion type
```
ericzolf, line 205: That's more generic python recommended practice, isn't it? Not wrong, but is it applicable to this document? Perhaps we should just point overall at some similar document of good practices, but for Python, and leave such topics out of here? This said, I don't know of such document, any idea welcome...

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+                    module.fail_json(msg="group %s will be automatically created by rule %s but "
+                                         "no description was provided" % (group_name, rule))
+                except ClientError as e:  # pylint: disable=duplicate-except
+                    module.fail_json_aws(e)
+----
+
+
+====
+
+== Plugins for applicance configuration should follow the Resource Module pattern (see below)
```
ericzolf, line 231: This one seems very very specific, and I don't really understand it. As it seems also that you're just pointing at existing documentation, I would skip it. We (IMHO) don't need to do deep-linking to the official documentation.

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+Explanations:: Resource modules allow the user to manage resources on an appliance (such as a network or security device) without having to write complex Jinja templates.
+
+Rationale:: Resource modules allow the user to manage resources on an appliance (such as a network or security device) without having to write complex Jinja templates.
+
+Examples::
+https://docs.ansible.com/ansible/latest/network/user_guide/network_resource_modules.html
+
+====
+
+== Use clear error/info messages
```
ericzolf, line 243: That's obvious developer good practices (but I can imagine that it remains a generic issue). Perhaps enlarge the topic to the module.v(v(v...) and other messaging functions (debug?), and of their good use? I know that I've never seen docs for those, and always need to read through different plugins to find out which ones they are.

ericzolf (changes requested):
A lot of comments, mostly on the form, not the content itself, on which you surely have much more knowledge.

As much as I expect people reading this document to have correct Ansible knowledge, I don't think we can expect as much knowledge when it comes to write plugins/modules, but I'm honest enough to say that this understanding is based on my own level of knowledge, so I'm happy to discuss.

And regarding the "should", I didn't write it anywhere, but it makes any recommendation weak, and I've had trainings on avoiding this way of phrasing things, so we... should avoid it.

plugins/README.adoc
jillr: @cidrblock I believe the jump host recommendation may have been yours.  How would you elaborate on this one?

plugins/README.adoc
jillr: Does it make sense to keep this as a generic recommendation?

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+====
+
+== Use sphinx (reST) formatted docstrings
+[%collapsible]
+====
+Explanations::
+All classes, private and public functions should have a sphinx (reST) formatted docstring including all parameters, yields, raises, or returns.
+
+Rationale::
+PEP-257 states that all modules should normally have docstrings, and all functions and classes exported by a module should also have docstrings. Public methods (including the __init__ constructor) should also have docstrings. A package may be documented in the module docstring of the __init__.py file in the package directory.
```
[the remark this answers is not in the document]
jillr: I'll need to keep one "should" as the text comes directly from the PEP, but I've made it more obvious that it's a quotation (and linked).

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+====
+
+== The use of unittest is discouraged, use pytest instead.
+[%collapsible]
+====
+Explanations:: Use pytest for writing unit tests for plugins
+
+Rationale:: Pytest is the testing framework used by Ansible Engineering and will provide the best experience for plugin developers
+
+Examples::
```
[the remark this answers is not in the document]
jillr: We don't have anything in the docs that specifically discusses preferences between unit testing frameworks (and there are certainly many in use in the existing code).  The closest documentation I can link here is the general "how to unit test ansible modules" [doc](https://docs.ansible.com/ansible/latest/dev_guide/testing_units_modules.html) and I'll be the first to admit that it's not for the faint of heart.

We want to encourage plugin developers to write tests, but this is a more advanced developer topic.

Do you believe there would be value in keeping this type of recommendation here if we can expand appropriately on it?  Or would it be better to have a general recommendation for unit testing that links to the devel guide, without prescribing a particular framework?

plugins/README.adoc
jillr: @NilashishC @cidrblock Thinking of this PR in the context of general guidelines (not, for example, Network team coding practices or platform specific edge cases), what changes should we make here?  Or is the section OK as-is?

plugins/README.adoc
jillr: As we can't provide documentation or user-accessible examples I'm removing this recommendation.  It can be proposed again in the future if the circumstance changes.

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+From the https://github.com/ansible-collections/ansible.utils/blob/2.6.0/plugins/filter/ipaddr.py[ipaddr] filter plugin.
+[source,python]
+----
+from ansible_collections.ansible.utils.plugins.module_utils.common.argspec_validate import (
+    AnsibleArgSpecValidator,
+)
+
+
+aav = AnsibleArgSpecValidator(
+    data=data, schema=DOCUMENTATION, name="ipaddr"
```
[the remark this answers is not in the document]
jillr: I've removed the recommendation here as we don't seem to be able to support it with appropriate documentation or user guidance.

plugins/README.adoc
```diff
@@ -7,3 +7,256 @@ NOTE: Work in Progress...
...
+@pytest.mark.parametrize('directory, expected', ((d[0], d[2]) for d in ONE_DIR_DATA))
+def test_split_pre_existing_dir_one_level_exists(directory, expected, mocker):
+    mocker.patch('os.path.exists', side_effect=[True, False, False])
+    split_pre_existing_dir(directory) == expected
+
+----
+
+====
+
+== A complex plugin argspec should be derived from its documentation runtime if the plugin’s source is manually maintained.
```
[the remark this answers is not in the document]
jillr: As we can't provide documentation or user-accessible examples I'm removing this recommendation.  It can always be proposed again in the future if the circumstance changes.

plugins/README.adoc
jillr: Removing this section

plugins/README.adoc
jillr: I've added to this section. Thoughts?

plugins/README.adoc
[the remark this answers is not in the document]
ericzolf: These are all good questions to which I might not have a definitive answer:

1. plugin developement is definitely something we want to have in, and it's an advanced topic where I'm a rookie, so my level of knowledge might not be the bar to measure
2. a bit more rationale might be good, it would be super to understand why the developers use it, but just explaining that the documentation is built around this might be sufficient
3. a link to the official documentation would be of course good

plugins/README.adoc
```diff
@@ -7,3 +7,190 @@ NOTE: Work in Progress...
...
+
+
+# Display information only when a user has set an increased level of verbosity
+# ansible-playbook -i inventory -vvvv test-playbook.yml
+from ansible.utils.display import Display
+
+display = Display()
+
+lookupfile = self.find_file_in_search_path(variables, 'files', term)
+display.vvvv(u"File lookup using %s as file" % lookupfile)
```
ericzolf, line 192: Isn't the u-string a relict from Python2 or is it part of the compatibility python2to3?

jillr: It is, and Python 2.7 is still supported for target hosts.  Granted, I took this example from a lookup plugin and Py2 is only supported on the Controller for certain releases.  I can remove the unicode prefix though to encourage 3-syntax.

plugins/README.adoc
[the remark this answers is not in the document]
jillr: The dev guide has a whole Why, When, and How for unit tests so I'll note that info can be found there link out.

plugins/README.adoc
ericzolf: Yep, now I remember, the last thing was the Python2-ism :1st_place_medal:

ericzolf (approved):

ericzolf:
Approved in today's CoP meeting. Thanks, we really need someone who really knows their plugin writing (beyond copy -> paste -> adapt)

PR #64, opened 2022-05-18
Add recommendations for coding styles

jillr:
Draft of coding styles recommendations from ansible eng

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -27,7 +27,7 @@ Making all names valid identifiers will avoid encountering problems in the futur
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
-* Name all tasks, plays, and task blocks, and and start names with a capital letter, to improve readability.
+* Name all tasks, plays, and task blocks to improve readability.
 * Write task names in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
 * Avoid abbreviations in names, or use capital letter for abbreviations where it cannot be avoided.
 
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -27,9 +27,8 @@ Making all names valid identifiers will avoid encountering problems in the futur
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
-* Name all tasks and and start names with a capital letter, this improves readability.
+* Name all tasks, plays, and task blocks, and and start names with a capital letter, to improve readability.
 * Write task names in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
-* Name all plays and task blocks.
 * Avoid abbreviations in names, or use capital letter for abbreviations where it cannot be avoided.
 
 == YAML and Jinja2 Syntax
@@ -174,12 +173,13 @@ Task files that are very or and/or contain highly nested blocks are difficult to
 Breaking a large or complex task file into multiple discrete files makes it easier to read and understand what is being done in each part.
 ====
 
-* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
+* Use bracket notation instead of dot notation for value retrieval (e.g. `item['key']` vs. `item.key`)
 +
 [%collapsible]
 ====
 Rationale::
 Dot notation will fail in some cases (such as when a variable name includes a hyphen) and it's better to stay consistent than to switch between the two options within a role or playbook.
+Additionally, some key names collide with attributes and methods of Python dictionaries such as `count`, `copy`, `title`, and others (refer to the https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#referencing-key-value-dictionary-variables[Ansible User Guide] for an extended list)
 
 Example::
 This https://blog.networktocode.com/post/Exploring-Jinja-Variable-Syntax-in-Ansible[post] provdes an excellent demonstration of how using dot notation syntax can impact your playbooks.
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -23,15 +23,14 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
 Making all names valid identifiers will avoid encountering problems in the future. Dictionary keys that are not valid identifiers are also less intuitive to use in Jinja2 (a dot in a dictionary key would be particularly confusing).
 ====
 
-* Use human-readable variable names
-* Use mnemonic and descriptive names and do not shorten more than necessary.
+* Use mnemonic and descriptive names that are human-readable and do not shorten more than necessary.
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
-* All tasks should be named and start with a capital letter, this improves readability.
-* All task names should be written in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
+* Name all tasks and and start names with a capital letter, this improves readability.
+* Write task names in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
 * Name all plays and task blocks.
-* Avoid abbreviations in names. or use capital letter for abbreviations where it cannot be avoided.
+* Avoid abbreviations in names, or use capital letter for abbreviations where it cannot be avoided.
 
 == YAML and Jinja2 Syntax
 
@@ -166,15 +165,24 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
   your task will likely need options such as `changed_when:` and maybe `check_mode:`).
 * Anytime `command` or `shell` modules are used, add a comment in the code with justification to help with future maintenance.
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
-* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
-* The maximum block depth within a single task file should be 2
+* Break complex task files down into discrete parts.
++
+[%collapsible]
+====
+Rationale::
+Task files that are very or and/or contain highly nested blocks are difficult to maintain.
+Breaking a large or complex task file into multiple discrete files makes it easier to read and understand what is being done in each part.
+====
+
 * Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
 +
 [%collapsible]
 ====
 Rationale::
 Dot notation will fail in some cases (such as when a variable name includes a hyphen) and it's better to stay consistent than to switch between the two options within a role or playbook.
-For an example, see https://blog.networktocode.com/post/Exploring-Jinja-Variable-Syntax-in-Ansible
+
+Example::
+This https://blog.networktocode.com/post/Exploring-Jinja-Variable-Syntax-in-Ansible[post] provdes an excellent demonstration of how using dot notation syntax can impact your playbooks.
 ====
 
 * Do not use `meta: end_play`.
@@ -271,4 +279,3 @@ Grouping sets of similar files into a subdirectory of `templates` is allowable, 
 * Using agnostic modules like `package` only makes sense if the features required are very limited.
   In many cases, if the platform is different, the package name is also different so that using `package` doesn't help a lot.
   Prefer then the more specific `yum`, `dnf` or `apt` module if you anyway need to differentiate.
-
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -23,10 +23,14 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
 Making all names valid identifiers will avoid encountering problems in the future. Dictionary keys that are not valid identifiers are also less intuitive to use in Jinja2 (a dot in a dictionary key would be particularly confusing).
 ====
 
+* Use human-readable variable names
 * Use mnemonic and descriptive names and do not shorten more than necessary.
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
+* All tasks should be named and start with a capital letter, this improves readability.
+* All task names should be written in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
+* Name all plays and task blocks
 
 == YAML and Jinja2 Syntax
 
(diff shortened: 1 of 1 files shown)


jillr:
cc @cidrblock

thedoubl3j (approved):

coding_style/README.adoc
```diff
@@ -162,6 +166,11 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
 * Anytime `command` or `shell` modules are used, add a comment in the code with justification to help with future maintenance.
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
 * Do not use `meta: end_play`.
+* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
+* The maximum block depth within a single task file should be 2
+* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
+
+
 +
 [%collapsible]
```
MarkusTeufelberger, line 175: This section belongs above the added section since it is part of the "Do not use `meta: end_play`." rule.

jillr: Thank you, good catch!

coding_style/README.adoc
```diff
@@ -161,6 +165,9 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
   your task will likely need options such as `changed_when:` and maybe `check_mode:`).
 * Anytime `command` or `shell` modules are used, add a comment in the code with justification to help with future maintenance.
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
+* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
+* The maximum block depth within a single task file should be 2
+* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
```
NilashishC, line 170: I am curious about what benefit the former has over the later. Is it somehow faster to evaluate? Or it's just easier to read?

ssbarnea (approved):

coding_style/README.adoc
cidrblock: the bracket vs. dot?  There are cases where dot notation can't be used, this rule request links to a good doc about it from NTC: https://github.com/ansible/ansible-lint/issues/2174

It's about consistency more than anything (for lab/home/non-published playbooks I still use dot)

https://github.com/ansible/ansible-lint/issues/2174

coding_style/README.adoc
```diff
@@ -161,6 +165,9 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
   your task will likely need options such as `changed_when:` and maybe `check_mode:`).
 * Anytime `command` or `shell` modules are used, add a comment in the code with justification to help with future maintenance.
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
+* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
```
ganeshrn, line 168: Any explanation of how we  came to the number 30 tasks? Based on how the role is implemented 30 tasks can be written in less than 100 lines and on the other side 5 tasks can take more than 500 lines (eg. network tasks)

coding_style/README.adoc
```diff
@@ -23,10 +23,14 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
...
 ====
 
+* Use human-readable variable names
 * Use mnemonic and descriptive names and do not shorten more than necessary.
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
+* All tasks should be named and start with a capital letter, this improves readability.
+* All task names should be written in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
+* Name all plays and task blocks
```
ganeshrn, line 33: May be add a comment to avoid abbreviations in name and/or use capital letter for abbreviations where it cannot be avoided

coding_style/README.adoc
sean-m-sullivan: I think its arbitrary, but the idea being is that a role could have 1000 tasks, Just that the task files should be broken down into manageable files for different tasks.

sean-m-sullivan (approved):
LGTM, its going to evolve over time, but this is a good start.

coding_style/README.adoc
jillr: I believe it was an arbitrary value, this is taken from in the internal draft from last month.  I don't have especially strong feelings about exactly what number goes here.

priyamsahoo (approved):

coding_style/README.adoc
```diff
@@ -23,10 +23,15 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
 Making all names valid identifiers will avoid encountering problems in the future. Dictionary keys that are not valid identifiers are also less intuitive to use in Jinja2 (a dot in a dictionary key would be particularly confusing).
 ====
 
+* Use human-readable variable names
```
ericzolf, line 26: I think that's a duplicate of the next bullet, either skip it or somehow merge it.

coding_style/README.adoc
```diff
@@ -23,10 +23,15 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
...
 * Use mnemonic and descriptive names and do not shorten more than necessary.
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
+* All tasks should be named and start with a capital letter, this improves readability.
```
ericzolf, line 31: I disagree with this one, I name my tasks with a non-capital letter, it looks IMHO better in the resulting log.

This said, I'm flabbergasted that I oversaw the rule to have all tasks named, so well catched.

coding_style/README.adoc
```diff
@@ -23,10 +23,15 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
...
+* All task names should be written in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
```
ericzolf, line 32: No should :-)

coding_style/README.adoc
```diff
@@ -23,10 +23,15 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
...
+* Name all plays and task blocks.
+* Avoid abbreviations in names. or use capital letter for abbreviations where it cannot be avoided.
```
ericzolf, line 34:
```suggestion
* Avoid abbreviations in names or use capital letter for abbreviations where it cannot be avoided.
```
A bit too generic for my taste, but not bad either.

coding_style/README.adoc
ericzolf: Few comments on this one:
- I don't think we want a concrete number, just a principle, each team can decide what works for them.
- This statement does belong here, it's more a role than code topic, but actually I think we have it already covered in https://github.com/redhat-cop/automation-good-practices/blob/main/structures/README.adoc (the "component" rationale, splitting big tasks into manageable bits).

coding_style/README.adoc
```diff
@@ -161,6 +166,17 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
+* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
+* The maximum block depth within a single task file should be 2
+* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
++
+[%collapsible]
+====
+Rationale::
+Dot notation will fail in some cases (such as when a variable name includes a hyphen) and it's better to stay consistent than to switch between the two options within a role or playbook.
+For an example, see https://blog.networktocode.com/post/Exploring-Jinja-Variable-Syntax-in-Ansible
```
ericzolf, line 177:
```suggestion
Example:: see https://blog.networktocode.com/post/Exploring-Jinja-Variable-Syntax-in-Ansible
```

coding_style/README.adoc
```diff
@@ -255,3 +271,4 @@ Grouping sets of similar files into a subdirectory of `templates` is allowable,
 * Using agnostic modules like `package` only makes sense if the features required are very limited.
   In many cases, if the platform is different, the package name is also different so that using `package` doesn't help a lot.
   Prefer then the more specific `yum`, `dnf` or `apt` module if you anyway need to differentiate.
+
```
ericzolf, line 274: why an empty line?

coding_style/README.adoc
```diff
@@ -161,6 +166,17 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
   your task will likely need options such as `changed_when:` and maybe `check_mode:`).
 * Anytime `command` or `shell` modules are used, add a comment in the code with justification to help with future maintenance.
 * Use the `| bool` filter when using bare variables (expressions consisting of just one variable reference without any operator) in `when`.
+* No single task file should contain more than 30 tasks.  Complex task files should be broken down into discrete parts.
+* The maximum block depth within a single task file should be 2
```
ericzolf, line 170: No should, avoid concrete number, something "Limit the block depth within a single task file for readability reasons", and 2 can be part of the rationale (experience shows...) or the example.

ericzolf (changes requested):
Again good inputs, more formal comments than regarding the content.

Overall we're trying to align this section (and the roles one) to the overall pattern explanation/rationale/example, so if you could start to stick to it, that would be helpful, but on the other hand, the coding recommendations are generally very short so it's a lot of boilerplate...

Regarding the task names, I really hate having them start with a capital letter (only my play names start with a capital letter), but we can discuss this in the group.

coding_style/README.adoc
jillr: rfc2119 sneaking in here - thanks for catching all the places I missed using the repo style! :)

coding_style/README.adoc
MarkusTeufelberger: I personally prefer concrete numbers. You can choose to change/ignore them, but it makes it easier to set an actual numerical limit rather than a "this file feels a bit too long" type of argument.

Maybe put a limit on number of lines in a file instead of/in addition to number of tasks? They can range from 2 lines (name, ansible.builtin.command: foo) to very long invocations with lots of parameters and so on.

coding_style/README.adoc
tima: Why 30? That seems arbitrary without an explanation to why specifically 30. What is meant by "discrete parts"? I have sophisticated automation play with 50 tasks. Is the 30 task an include to another task file that contains the other 20?

ssbarnea: Because it must be a number. Why black has a line limit of 88. They had to make a decision.

TBH, I do think that 30 is reasonable and we will find-out how big is the impact while we have a PR to implement it in linter, so we could tune it (before merging). 20 seems dangerously low for me, while 50 likely too big. Maybe setting it to "42", to play safe? ;)

Keep in mind that users might be able to disable the rule per file. Still, I would be (softly) against adding a configurable option.

coding_style/README.adoc
```diff
@@ -161,6 +165,26 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
+* Break complex task files down into discrete parts.
++
+[%collapsible]
+====
+Rationale::
+Task files that are very or and/or contain highly nested blocks are difficult to maintain.
+Breaking a large or complex task file into multiple discrete files makes it easier to read and understand what is being done in each part.
+====
+
+* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
```
mophahr, line 177:
```suggestion
* Use bracket notation instead of dot notation for value retrieval (e.g. `item['key']` vs. `item.key`) 
```

coding_style/README.adoc
```diff
@@ -23,10 +23,14 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
 Making all names valid identifiers will avoid encountering problems in the future. Dictionary keys that are not valid identifiers are also less intuitive to use in Jinja2 (a dot in a dictionary key would be particularly confusing).
 ====
 
-* Use mnemonic and descriptive names and do not shorten more than necessary.
+* Use mnemonic and descriptive names that are human-readable and do not shorten more than necessary.
   A pattern `object[_feature]_action` has proven useful as it guarantees a proper sorting in the file system for roles and playbooks.
   Systems support long identifier names, so use them!
 * Avoid numbering roles and playbooks, you'll never know how they'll be used in the future.
+* Name all tasks and and start names with a capital letter, this improves readability.
```
mophahr, line 30:
```suggestion
* Name all tasks, plays, and task blocks, and start names with a capital letter to improve readability.
```

coding_style/README.adoc
```diff
@@ -23,10 +23,14 @@ Rationale:: even when Ansible currently allows names that are not valid identifi
...
+* Write task names in the imperative (e.g. "Ensure service is running"), this communicates the action of the task.
+* Name all plays and task blocks.
```
mophahr, line 32:
```suggestion
```

coding_style/README.adoc
```diff
@@ -161,6 +165,26 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
+Task files that are very or and/or contain highly nested blocks are difficult to maintain.
+Breaking a large or complex task file into multiple discrete files makes it easier to read and understand what is being done in each part.
+====
+
+* Bracket notation is preferred over dot notation for value retrieval (e.g. `item[‘key’]` vs. `item.key`)
++
+[%collapsible]
+====
+Rationale::
+Dot notation will fail in some cases (such as when a variable name includes a hyphen) and it's better to stay consistent than to switch between the two options within a role or playbook.
```
mophahr, line 182:
```suggestion
Dot notation will fail in some cases (such as when a variable name includes a hyphen) and it's better to stay consistent than to switch between the two options within a role or playbook.
Additionally, some key names collide with attributes and methods of Python dictionaries such as `count`, `copy`, `title`, and others (refer to the https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#referencing-key-value-dictionary-variables[Ansible User Guide] for an extended list)
```

mophahr (changes requested):
I have a few small suggestions but otherwise this looks good to me.

coding_style/README.adoc
jillr: @ericzolf Do you consider this line a blocker for this PR?  I'm not clear if you have a differing opinion but don't feel strongly or if we need to remove this guidance.

ericzolf: I'd be fine with "Name all tasks to improve readability of the logs", I've created a separate issue to discuss capital or non-capital names (we can also decide to _not_ make a recommendation).

bontreger (approved):

coding_style/README.adoc
ssbarnea: @ericzolf WDYT about opening threads under https://github.com/ansible/ansible-lint/discussion even as "polls" is we already know the possible directions or normal ideas if they directions are not yet clear. This will makes the debate/decision process considerably more inclusive?

In fact there are several checks we need to implement and decide on related to name content: capitalized first letter? should it be imperative? should it end with a dot or not? should we ensure that same task name does not repeat inside current file, current role, globally? I did talk with @cidrblock and we are likely going to try to re-use python docstring checks for this.

ericzolf (approved):

PR #65, opened 2022-05-23
Blocked by #61: Add recommendations for roles

jillr:
Draft of roles recommendations from ansible eng

cc @cidrblock

### DIFF
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -386,7 +386,7 @@ At minimum include:
 * Example playbooks that allow users to understand how the developed content works are also part of the documentation.
 * The inbound and outbound role argument specifications
 * List of user-facing capabilities within the role
-* The unit of automation the role provided
+* The unit of automation the role provides
 * The outcome of the role
 * The roll-back capabilities of the role
 * Designation of the role as idempotent (True/False)
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -460,13 +460,6 @@ The log output will then look something like `TASK [myrole : sub | some task des
 TIP: with a verbosity of 2 or more, ansible-playbook will show the full path to the task file, but this generally means that you need to restart the play in a higher verbosity to get the information you could have had readily available.
 ====
 
-=== License Documentation
-[%collapsible]
-* Place the license file in the root directory
-* Name the file LICENSE or COPYING. The contents must be the text of the applicable license, or a link to the canonical reference for the license on the Internet (such as https://opensource.org/licenses/BSD-2-Clause)
-* Any file that is licensed differently from the larger collection it is a part of (such as module utilities) must have that license noted in the header of the file.
-
-
 == References
 [%collapsible]
 ====
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -373,7 +373,7 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
 * Use RFC https://tools.ietf.org/html/rfc5737[5737], https://tools.ietf.org/html/rfc7042#section-2.1.1[7042] and https://tools.ietf.org/html/rfc3849[3849] addresses in examples.
 ====
 
-=== README file
+=== Create a meaningful README file for every role
 [%collapsible]
 ====
 Rationale::
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -382,6 +382,7 @@ Place the README file in the root directory of the role.
 The README file exists to introduce the user to the purpose of the role and any important information on how to use it, such as credentials that are required.
 
 At minimum include:
+
 * Example playbooks that allow users to understand how the developed content works are also part of the documentation.
 * The inbound and outbound role argument specifications
 * List of user-facing capabilities within the role
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -98,6 +98,7 @@ Consider making paths to templates internal variables if you need different temp
 
 === Package roles in an Ansible collection to simplify distribution and consumption
 [%collapsible]
+====
 Rationale::
 Packaging roles as a collection allows you to distribute many roles in a single cohesive unit of re-usable automation.
 Inside a collection, you can share custom plugins across all roles in the collection instead of duplicating them in each role’s library/` directory.
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -94,6 +94,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
 Since distribution-specific values are kind of "magic values", it applies to them.
 The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
 Consider making paths to templates internal variables if you need different templates for different distributions.
+A role should have a single entry point. Requiring users to use `tasks_from: file.yml` is discouraged as this could cause important pre and post task files executed in `tasks/main.yml` to be skipped.
+The `state` or `action` keyword should be passed to a role when a role has such a capability and be required when calling the role. The `tasks/main.yml` should load the appropriate task file based on the action or state requested by the user, rather than the consumer having to have intimate knowledge of the file structure layout within the role.
+All facts added by a role should fall under a single fact key named as the role.
+All roles should use argspec validation, if multiple states or actions are supported argspec validation should be done with a task.
+All roles should use argspec validation at the egress to confirm facts are correctly set.
+The tasks folder or one of its subdirectories should not contain more than 10 files.
+All intermediate facts set during role execution should be prefaced with the role name.
+All roles should be packaged within an ansible collection
+Configuration management roles should be atomic, idempotent, and rollback when failures are encountered
+When many roles within a collection are similar, consider using a dispatcher role as the single entry point to invoke a specific role within the collection. An example of such may be a collection that does system health checks and contains many roles, each representing an individual health check.
 ====
 
 === Check Mode
@@ -360,7 +370,29 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
 * Use fully qualified role names in examples, like: `linux-system-roles.$ROLENAME` (with the Galaxy prefix).
 * Use RFC https://tools.ietf.org/html/rfc5737[5737], https://tools.ietf.org/html/rfc7042#section-2.1.1[7042] and https://tools.ietf.org/html/rfc3849[3849] addresses in examples.
 * Modules should have complete metadata, documentation, example and return blocks as described in the https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html[Ansible docs].
+* All roles should have a complete README and meta/main.yml
+* All entries in a defaults/main.yml should be documented with comments
+* The ingress and egress argspec should be programmatically included in the README
 ====
+
+=== README file
+[%collapsible]
+====
+Rationale::
+The documentation is essential for the success of the content.
+A README file should be included in the root directory of the role that introduces the user to the purpose of the role and any important information on how to use it, such as credentials that are required.
+The role’s README should be programmatically maintained and kept up to date.
+
+Should include at least:
+* Example playbooks that allow users to understand how the developed content works are also part of the documentation.
+* The inbound and outbound role argument specifications
+* List of user-facing capabilities within the role
+* The unit of automation the role provided
+* The outcome of the role
+* The roll-back capabilities of the role
+* Designation of the role as idempotent (True/False)
+* Designation of the role as atomic if applicable (True/False)
+=====
 
 === Don't use host group names or at least make them a parameter
 [%collapsible]
@@ -430,6 +462,13 @@ The log output will then look something like `TASK [myrole : sub | some task des
 +
 TIP: with a verbosity of 2 or more, ansible-playbook will show the full path to the task file, but this generally means that you need to restart the play in a higher verbosity to get the information you could have had readily available.
 ====
+
+=== License Documentation
+[%collapsible]
+* The license file should be included in the root directory
+* The license file for the collection should be named LICENSE or COPYING. The contents should be the text of the applicable license, or a link to the canonical reference for the license on the Internet (such as https://opensource.org/licenses/BSD-2-Clause )
+* Any file that is licensed differently from the larger collection it is a part of (such as module utilities) should have that license noted in the header of the file.
+
 
 == References
 [%collapsible]
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -151,7 +151,7 @@ If you just want to ensure the user always uses `gather_facts: true`, and do not
 
 If it is desirable to use roles that require facts, but fact gathering is expensive, consider using a cache plugin https://docs.ansible.com/ansible/latest/collections/index_cache.html[List of Cache Plugins], and also consider running a periodic job on the controller to refresh the cache.
 
-==== Platform specific variables
+=== Platform specific variables
 [%collapsible]
 ====
 Explanations:: You normally use `vars/main.yml` (automatically included) to set variables used by your role.
@@ -390,7 +390,7 @@ At minimum include:
 * The roll-back capabilities of the role
 * Designation of the role as idempotent (True/False)
 * Designation of the role as atomic if applicable (True/False)
-=====
+====
 
 === Don't use host group names or at least make them a parameter
 [%collapsible]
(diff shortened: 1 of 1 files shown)


branic (approved):
LGTM

ericzolf:
I don't agree with everything but I need a bit more time (and effort) to give a fair review.

NilashishC (approved):

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
 Since distribution-specific values are kind of "magic values", it applies to them.
 The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
 Consider making paths to templates internal variables if you need different templates for different distributions.
+A role should have a single entry point. Requiring users to use `tasks_from: file.yml` is discouraged as this could cause important pre and post task files executed in `tasks/main.yml` to be skipped.
+The `state` or `action` keyword should be passed to a role when a role has such a capability and be required when calling the role. The `tasks/main.yml` should load the appropriate task file based on the action or state requested by the user, rather than the consumer having to have intimate knowledge of the file structure layout within the role.
+All facts added by a role should fall under a single fact key named as the role.
+All roles should use argspec validation, if multiple states or actions are supported argspec validation should be done with a task.
```
ganeshrn, line 119: The argspec validation can be done in role argspec itself using conditional for different states. If validation is required at task level maybe it's an indication that role is getting complicated and logic should be moved to a separate role?

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+All roles should use argspec validation at the egress to confirm facts are correctly set.
+The tasks folder or one of its subdirectories should not contain more than 10 files.
+All intermediate facts set during role execution should be prefaced with the role name.
+All roles should be packaged within an ansible collection
+Configuration management roles should be atomic, idempotent, and rollback when failures are encountered
```
ganeshrn, line 124: Should it also support dry-run capability?

cidrblock: Yes, good catch. check-mode/dry run should be supported.......... I'm assuming there will be cases were the endpoint system makes it impossible, I think it would be best to call this out as a limitation of the endpoint not an omission from the content

roles/README.adoc
```diff
@@ -389,8 +399,30 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
 * Use fully qualified role names in examples, like: `linux-system-roles.$ROLENAME` (with the Galaxy prefix).
 * Use RFC https://tools.ietf.org/html/rfc5737[5737], https://tools.ietf.org/html/rfc7042#section-2.1.1[7042] and https://tools.ietf.org/html/rfc3849[3849] addresses in examples.
 * Modules should have complete metadata, documentation, example and return blocks as described in the https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html[Ansible docs].
+* All roles should have a complete README and meta/main.yml
+* All entries in a defaults/main.yml should be documented with comments
+* The ingress and egress argspec should be programmatically included in the README
```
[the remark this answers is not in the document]
jillr: If anyone knows of any examples I would love to include them - I couldn't find any

priyamsahoo (approved):

ericzolf:
As another side comment, we should wait for PR #61 to be merged before modifying the roles section too much, or @bontreger will have to redo his PR yet again, and that wouldn't be nice.

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
 Since distribution-specific values are kind of "magic values", it applies to them.
 The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
 Consider making paths to templates internal variables if you need different templates for different distributions.
+A role should have a single entry point. Requiring users to use `tasks_from: file.yml` is discouraged as this could cause important pre and post task files executed in `tasks/main.yml` to be skipped.
```
mophahr, line 116:
```suggestion
A role should have a single entry point. 
Requiring users to use `tasks_from: file.yml` is discouraged as this could cause important pre and post task files executed in `tasks/main.yml` to be skipped.
```

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+The `state` or `action` keyword should be passed to a role when a role has such a capability and be required when calling the role. The `tasks/main.yml` should load the appropriate task file based on the action or state requested by the user, rather than the consumer having to have intimate knowledge of the file structure layout within the role.
```
mophahr, line 117:
```suggestion
The `state` or `action` keyword should be passed to a role when a role has such a capability and be required when calling the role. 
The `tasks/main.yml` should load the appropriate task file based on the action or state requested by the user, rather than the consumer having to have intimate knowledge of the file structure layout within the role.
```

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+All facts added by a role should fall under a single fact key named as the role.
```
mophahr, line 118: I'm not convinced we need this. Maybe I also misnuderstand the rule. Could you please give an example?

roles/README.adoc
mophahr, line 119: I have never done this in my roles, so I don't really understand what this is about. Couild you please be more specific and add an example?

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+All roles should use argspec validation, if multiple states or actions are supported argspec validation should be done with a task.
+All roles should use argspec validation at the egress to confirm facts are correctly set.
+The tasks folder or one of its subdirectories should not contain more than 10 files.
+All intermediate facts set during role execution should be prefaced with the role name.
```
mophahr, line 122: I believe we already have a rule for this. please double-check

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+All roles should be packaged within an ansible collection
```
mophahr, line 123:
```suggestion
All roles should be packaged within an Ansible collection.
```

This needs a bit more context.

roles/README.adoc
mophahr, line 124:
```suggestion
Configuration management roles should be atomic, idempotent, and rollback when failures are encountered.
```

This is _huge_; especially the rollback part. I think we should discuss it in teh CoP meeting

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+Configuration management roles should be atomic, idempotent, and rollback when failures are encountered
+When many roles within a collection are similar, consider using a dispatcher role as the single entry point to invoke a specific role within the collection. An example of such may be a collection that does system health checks and contains many roles, each representing an individual health check.
```
mophahr, line 125:
```suggestion
When many roles within a collection are similar, consider using a dispatcher role as the single entry point to invoke a specific role within the collection. 
An example of such may be a collection that does system health checks and contains many roles, each representing an individual health check.
```

roles/README.adoc
```diff
@@ -389,8 +399,30 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
 * Use fully qualified role names in examples, like: `linux-system-roles.$ROLENAME` (with the Galaxy prefix).
 * Use RFC https://tools.ietf.org/html/rfc5737[5737], https://tools.ietf.org/html/rfc7042#section-2.1.1[7042] and https://tools.ietf.org/html/rfc3849[3849] addresses in examples.
 * Modules should have complete metadata, documentation, example and return blocks as described in the https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html[Ansible docs].
+* All roles should have a complete README and meta/main.yml
```
mophahr, line 402: why require `meta/main.yml`?

mophahr (changes requested):
there's a few things that need clarification. See my individual comments.

Also, please have a look at the contribution guide (https://github.com/redhat-cop/automation-good-practices/blob/main/CONTRIBUTE.adoc)

ericzolf (changes requested):
Sorry, prepared my review and forgot to send it, some duplicates/contradicts with @mophahr

roles/README.adoc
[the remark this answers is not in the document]
jillr: @cidrblock @ganeshrn if we can't support the arg spec validation recommendations with documentation or functional/user-accessible sample code I'd like to remove them.  Any objections?

roles/README.adoc
```diff
@@ -113,6 +113,16 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
 Consider making paths to templates internal variables if you need different templates for different distributions.
+A role should have a single entry point. Requiring users to use `tasks_from: file.yml` is discouraged as this could cause important pre and post task files executed in `tasks/main.yml` to be skipped.
+The `state` or `action` keyword should be passed to a role when a role has such a capability and be required when calling the role. The `tasks/main.yml` should load the appropriate task file based on the action or state requested by the user, rather than the consumer having to have intimate knowledge of the file structure layout within the role.
+All facts added by a role should fall under a single fact key named as the role.
+All roles should use argspec validation, if multiple states or actions are supported argspec validation should be done with a task.
+All roles should use argspec validation at the egress to confirm facts are correctly set.
+The tasks folder or one of its subdirectories should not contain more than 10 files.
```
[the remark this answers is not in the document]
tima: Agreed. I don't feel we should be setting arbitrary limits like this.

roles/README.adoc
tima: Yes I agree. There be demons here.

roles/README.adoc
tima: This example and recommendation is a curious to me. Is this that common a scenario? If not I'd recommend we leave it out especially if a tool like ansible-lint cannot easily identify the pattern and make a recommendation. My overarching concern with these guidelines is that they will overwhelm a would-be Ansible content developer from getting their heads wrapped around all of these and discourages them from even attempting to create something. 

Are these health checks going to be potentially run as one off operations? If not, then I question the value creating individual roles. This seems counter to my long standing understanding of roles being coarse-grain use case driven units of Ansible content and not programming libraries. Individual health checks sounds too fine-grain unless there is a compelling reasons why each individual health check may be executed independently with their own distinct handlers, defaults and vars. When I hear "dispatcher role" and "many similar roles within one collection" I feel like we are talking about writing coding and I wonder if the person who wrote that content is a Java programmer.

roles/README.adoc
jillr: @ericzolf What I understand here is that we should remove this recommendation from this PR/section in favor of the work happening in PR 42.  Please let me know if I misunderstand and I can add it back in.

roles/README.adoc
jillr: @ssbarnea we need to be able to provide better documentation, examples, and rationale for suggestions like this.  I'm going to drop it from this PR, and it would need to be re-proposed as it's own change so that we don't drag out this PR if we want to keep it.

roles/README.adoc
```diff
@@ -460,6 +492,13 @@ The log output will then look something like `TASK [myrole : sub | some task des
 TIP: with a verbosity of 2 or more, ansible-playbook will show the full path to the task file, but this generally means that you need to restart the play in a higher verbosity to get the information you could have had readily available.
 ====
 
+=== License Documentation
```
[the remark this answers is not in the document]
jillr: Why would stand alone roles not be appropriately licensed?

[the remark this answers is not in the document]
ericzolf: Two main reasons:

1. we want to encourage distribution by collections, don't we? And licensing is a distribution topic, not so much a code topic, so (IMHO) the topic fits better in the collection section
2. the last bullet is explicitly relating to collections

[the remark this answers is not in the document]
jillr: I can follow that logic.  This info is in #66 for collections so I'll remove it from this PR. Thanks.

roles/README.adoc
```diff
@@ -96,6 +96,17 @@ The same logic applies for providers: a role can load a provider-specific variab
 Consider making paths to templates internal variables if you need different templates for different distributions.
 ====
 
+=== Package roles in an Ansible collection to simplify distribution and consumption
+[%collapsible]
```
ericzolf, line 100:
```suggestion
[%collapsible]
====
```

ericzolf: That must have been the issue with the pipeline.

roles/README.adoc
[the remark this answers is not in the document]
ericzolf: Good point, let's keep it here.

sean-m-sullivan (approved):
This looks good for a start, not sure what else to add.

roles/README.adoc
```diff
@@ -389,8 +399,30 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
...
+* The ingress and egress argspec should be programmatically included in the README
 ====
 
+=== README file
+[%collapsible]
+====
+Rationale::
+The documentation is essential for the success of the content.
+A README file should be included in the root directory of the role that introduces the user to the purpose of the role and any important information on how to use it, such as credentials that are required.
+The role’s README should be programmatically maintained and kept up to date.
```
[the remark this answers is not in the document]
sean-m-sullivan: Agreed, right now this is a pain point for controller configuration.

ericzolf:
@mophahr will redo a review of the PR, it's easier as it's not that long.

@jillr you still need to address my suggestion at https://github.com/redhat-cop/automation-good-practices/pull/65/files#r915580222 because else the pipeline will fail.

roles/README.adoc
```diff
@@ -359,7 +371,25 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
...
+
+=== README file
+[%collapsible]
+====
+Rationale::
+The documentation is essential for the success of the content.
+Place the README file in the root directory of the role.
+The README file exists to introduce the user to the purpose of the role and any important information on how to use it, such as credentials that are required.
+
+At minimum include:
```
mophahr, line 384: the bullet points here aren't rendered correctly. maybe this works:

```suggestion
At minimum include:

```

roles/README.adoc
```diff
@@ -359,7 +371,25 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
 ====
 * Use fully qualified role names in examples, like: `linux-system-roles.$ROLENAME` (with the Galaxy prefix).
 * Use RFC https://tools.ietf.org/html/rfc5737[5737], https://tools.ietf.org/html/rfc7042#section-2.1.1[7042] and https://tools.ietf.org/html/rfc3849[3849] addresses in examples.
-* Modules should have complete metadata, documentation, example and return blocks as described in the https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html[Ansible docs].
+====
+
+=== README file
```
mophahr, line 376:
```suggestion
=== Create a meaningful README file for every role
```

roles/README.adoc
```diff
@@ -359,7 +371,25 @@ See https://docs.ansible.com/ansible/latest/dev_guide/overview_architecture.html
...
+Rationale::
+The documentation is essential for the success of the content.
+Place the README file in the root directory of the role.
+The README file exists to introduce the user to the purpose of the role and any important information on how to use it, such as credentials that are required.
+
+At minimum include:
+* Example playbooks that allow users to understand how the developed content works are also part of the documentation.
+* The inbound and outbound role argument specifications
+* List of user-facing capabilities within the role
+* The unit of automation the role provided
```
mophahr, line 388: I'm not sure if I get what a "unit of automation is", could you please expand this a bit?

```suggestion
* The unit of automation the role provides
```

mophahr (changes requested):
As announced by @ericzolf, I've reviewed the content. 

@jillr I have made some new small aesthetic suggestions you could apply. Once adressed, i think we can merge this.

@ericzolf could you please either approve or retract your "needs changes" on this PR?

mophahr:
Thanks for the changes @jillr ! Approved from my side.

@ericzolf please remove/change your review.

mophahr (approved):

ericzolf:
Approved in CoP meeting today. Again thanks for your patience and good work!

PR #67, opened 2022-05-27
Clarify why not to use eq, equalto, and ==

richm:
Add some clarification about why the use of `eq`, `equalto`, and `==`
are discouraged, and how to use `match`, `search`, and `regex` instead.

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -187,7 +187,34 @@ A fact has a different priority than other variables and not the highest, so in 
   Role variables are exposed to the whole play when the role is applied using `roles:` or `import_role:`. A more restricted scope such as task or block variables is preferred.
 * Beware of `ignore_errors: true`; especially in tests.
   If you set on a block, it will ignore all the asserts in the block ultimately making them pointless.
-* Do not use the `eq` (introduced in Jinja 2.10) or `equalto` Jinja operators.
+* Do not use the `eq`, `equalto`, or `==` Jinja tests introduced in Jinja 2.10, use Ansible built-in `match`, `search`, or `regex` instead.
++
+[%collapsible]
+====
+Rationale:: These tests are not present in older versions of Jinja used on older controller platforms.
+If you want to ensure that your code works on older platforms, use the built-in Ansible tests such as (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[match]), (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[search]), or (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[regex]) instead.
+For example, if you have a `list` of `dict`, and you want to filter out elements that have the key `type` with the value `bad_type`.
+
+.Do this:
+[source,yaml]
+----
+tasks:
+  - name: Do something
+    some.module:
+      param: "{{ list_of_dict | rejectattr('type', 'search', '^bad_type$') | list }}"
+----
+
+.Don't do this:
+[source,yaml]
+----
+tasks:
+  - name: Do something
+    some.module:
+      param: "{{ list_of_dict | rejectattr('type', 'eq', 'bad_type') | list }}"
+----
+When using `match`, `search`, or `regex`, and you want an exact match, you must specify the regex `^STRING$`, otherwise, you will match partial strings.
+====
+
 * Avoid the use of `when: foo_result is changed` whenever possible.
   Use handlers, and, if necessary, handler chains to achieve this same result.
 * Use the various include/import statements in Ansible.


branic (approved):
LGTM

coding_style/README.adoc
```diff
@@ -187,7 +187,34 @@ A fact has a different priority than other variables and not the highest, so in
...
 * Beware of `ignore_errors: true`; especially in tests.
   If you set on a block, it will ignore all the asserts in the block ultimately making them pointless.
-* Do not use the `eq` (introduced in Jinja 2.10) or `equalto` Jinja operators.
+* Do not use the `eq`, `equalto`, or `==` Jinja tests introduced in Jinja 2.10, use Ansible built-in `match`, `search`, or `regex` instead.
++
+[%collapsible]
+====
+Rationale:: These tests are not present in older versions of Jinja used on older controller platforms.
+If you want to ensure that your code works on older platforms, use the built-in Ansible tests such as (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[match]), (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[search]), or (https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#testing-strings[regex]) instead.
+For example, if you have a `list` of `dict`, and you want to filter out elements that have the key `type` with the value `bad_type`.
```
ericzolf, line 196:
```suggestion
Example::
if you have a `list` of `dict`, and you want to filter out elements that have the key `type` with the value `bad_type`.
```

coding_style/README.adoc
```diff
@@ -187,7 +187,34 @@ A fact has a different priority than other variables and not the highest, so in
   Role variables are exposed to the whole play when the role is applied using `roles:` or `import_role:`. A more restricted scope such as task or block variables is preferred.
 * Beware of `ignore_errors: true`; especially in tests.
   If you set on a block, it will ignore all the asserts in the block ultimately making them pointless.
-* Do not use the `eq` (introduced in Jinja 2.10) or `equalto` Jinja operators.
+* Do not use the `eq`, `equalto`, or `==` Jinja tests introduced in Jinja 2.10, use Ansible built-in `match`, `search`, or `regex` instead.
```
ericzolf, line 190: What I don't like about this one, is that it's time-bombed: once Jinja 2.10 becomes common, it's obsolete and might even make code unnecessarily complex.

Any idea how we could handle generally this kind of recommendations? Add a :bomb: in the title and explain what it means in the introduction of the document?

ericzolf (changes requested):

coding_style/README.adoc
richm: > I don't think we would need the bomb often enoughj to warrant introducing it ;)
> 
> But I'm still confused as to which ansible versions are affected by this.

Depends on what you mean by "Ansible", if you mean "Ansible and the underlying OS platform libraries".  The issue is only with Jinja versions older than 2.10.  RPM distributions of Ansible generally use the underlying OS platform python library for jinja e.g. python-jinja2.  This is especially problematic on EL7.  The only supported Ansible RPM is 2.9, which uses the EL7 platform python-jinja2 library, which is 2.7 (and will likely never be upgraded).  We still have a _lot_ of customers using EL7 for the control node.  I believe this means AAP 1.x customers will also be affected.  Customers not affected:
* AAP 2.x customers - I believe there is an option to use EL8 runners, or otherwise, build the EEs in such a way as to use jinja 2.11 or later
* Customers running Ansible from a pip install (but these are unsupported)
* Customers running Ansible installed via RPM on EL8 or later

ericzolf: > I don't think we would need the bomb often enough to warrant introducing it ;)

Sounds like famous last words to me. It's the first one, but I guess, the older RHEL 7 becomes, the more Richard, or others, will have similar requests. I'm thinking that perhaps we should create a new category "tips & tricks".

At the end it's more a "how to keep compatibility with older versions of Jinja2" and _not_ "you need to keep compatibility with older versions of Jinja", because this would be a product decision, not a generic recommendation (if you don't care about RHEL7, what's the point?). Probably something to discuss in the bi-weekly meeting?

ericzolf: @richm independent from my previous comment, your explanations about versions impacted could be of interest for the example section.

richm:
@ericzolf check out the latest version

mophahr (approved):

ericzolf (approved):

ericzolf:
I don't want to stand in the way of this change, but I still think that it's misplaced hence I created a new issue to discuss the topic #69

ericzolf:
Approved in CoP meeting 2022-06-29, will most probably to the newly agreed Tips & Tricks section, once created.

PR #68, opened 2022-05-31
Recommendations for Jinja and roles

jillr:
Add Jinja recommendations and some role text that was missed in previous PR.

This will likely conflict with 64,65; my expectation would be that those need to land first so I can rebase and clean this one up.


coding_style/README.adoc
```diff
@@ -150,6 +150,16 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
   But do use quotes for user-side strings such as descriptions, names, and messages.
 * Even if JSON is valid YAML and Ansible understands it, do only use JSON syntax if it makes sense (e.g. a variable file automatically generated) or adds to the readability.
   In doubt, nobody expects JSON so stick to YAML.
+* A single jinja template should not exceed 250 lines in length
```
ericzolf, line 153: I would create one single recommendation about keeping Jinja2 structures simple enough, and then have the bullet points become part of the explanation.

ericzolf: On this specific line:
1. I'm still struggling with the specific number
2. nobody creates a 250 lines Jinja template for fun (well, obfuscated Ansible contest perhaps?), I can't even imagine a good use case. But if someone would come with a reasonable use case, what would be the alternative approach?

coding_style/README.adoc
```diff
@@ -150,6 +150,16 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
+* The aggregate number of jinja includes should not exceed 3
```
ericzolf, line 154: What do you call aggregate number?

coding_style/README.adoc
```diff
@@ -150,6 +150,16 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
+* Jinja templates should not be used to create structured data but instead text and semi-structured data
```
ericzolf, line 155: Why not, I'm pretty sure, I've already done this and didn't feel especially bad about it.

coding_style/README.adoc
```diff
@@ -150,6 +150,16 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
+* Filter plugins are preferred over the use of data manipulation or transformation in Jinja templates when the necessary data transformation cannot be accomplished with existing and available filter plugins or the transformation of data is unique to the use case.
```
ericzolf, line 156: The sentence is difficult to understand, I think you mean actually two things:

1. prefer filter plugins to Jinja2 templates
2. for more complex use cases, write a custom filter plugin

Anyway you meant it, split the sentence somehow, I think it'll help.

coding_style/README.adoc
```diff
@@ -150,6 +150,16 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
++
++
+[%collapsible]
+====
+Rationale:: Long and complex Jinja templates can be difficult to maintain and debug. When working with structured data and/or data transformations it is preferable to use a programming language (such as Python) that has better support and tooling to do this kind of work. Additionally, tasks will be much more legible if data is managed and manipulated via plugins than with in-line Jinja.
```
ericzolf, line 161: I fully agree with the statement. Just one formatting thing: one line per sentence, one sentence per line.
https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

roles/README.adoc
```diff
@@ -113,6 +113,8 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
 Since distribution-specific values are kind of "magic values", it applies to them.
 The same logic applies for providers: a role can load a provider-specific variable file, include a provider-specific task file, or both, as needed.
 Consider making paths to templates internal variables if you need different templates for different distributions.
+Do not use roles as programming interfaces, in the case a common role is necessary to contain common code available to other roles within the same collection, it should not contain user-facing plugins.
```
ericzolf, line 116: I don't grok exactly what you're trying to say.

roles/README.adoc
```diff
@@ -113,6 +113,8 @@ See also <<vars-vs-defaults,Vars vs Defaults>> which mandates "Avoid embedding l
...
+Keep roles loosely coupled with other external roles and variables.
```
ericzolf, line 117: I think you need to be more precise. Actually it sounds to me like a good heading for other more concrete rules we've already stated.

roles/README.adoc
```diff
@@ -53,7 +53,7 @@ What should a role do and how can a user tell it what to do.
 === Basic design
 [%collapsible]
 ====
-Try to design the interface focused on the functionality, not on the software implementation behind it.
+Try to design the interface focused on the functionality, not on the software implementation behind it. Roles should accomplish a guaranteed outcome; the scope of a role should be its outcome
```
ericzolf, line 56: That sounds a bit fluffy, not wrong but difficult to apply practically.

ericzolf (changes requested):
We ask people to split their PR with one recommendation per PR, we're not too strict, but at least don't create PR across two sections, as it makes approval more difficult.
Especially because the roles section is being refactored so it's a bit stuck for now.

coding_style/README.adoc
tima: Agreed @ericzolf. This is another case of an arbitrary number I am opposed to having in these guidelines. If I have a large configuration file example from an ISV I need to template I should have to cut it up into piece just to reassemble it using includes order to comply here. All of these guidelines should focus the most common cases on providing a baseline that can be checked by tooling like ansible-lint.

coding_style/README.adoc
tima: Agreed.

roles/README.adoc
tima: I agree what this is aying isn't very clear.

coding_style/README.adoc
tima: Yeah this combined with the 250 line limit above seems arbitrary and potentially a pain.

ericzolf:
You should now be un-blocked!

coding_style/README.adoc
jillr: +1  We also discussed in another PR and I'll remove all recommendations for hard numbers

coding_style/README.adoc
jillr: While it can be done, Python provides far more tools for creating and manipulating data than a templating language.  Complex Jinja statements are especially challenging for less experienced users and are not especially readable.

Recent example:
I was helping someone troubleshoot a playbook they were working on.  They wanted to build a dict that used vars for some keys and some values.  In Python this sort of thing is trivial.  But since they were doing it in a playbook I had to walk them through understanding how to 1. build a dict in Jinja in the first place 2. write the template so that it would go through the Jinja parser before the YaML parser 3. how to include both vars and string literals where there were multiple special characters involved.  And this wasn't even that complicated of an example.

roles/README.adoc
cidrblock: I think we might be better to provide some "do do this" rather than a "dont' do this....

Some things that come to mind:

1) Strive to author loosely coupled hierarchical content, ravioli vs. spaghetti or lasagna (re pasta theory)

2) Place content common to multiple roles in a single role within the collection, "common" is a common name for this role.

3) Limit the consumer's need to understand specific implementation details about a collection, the role names within, and the file names. Presenting the collection as a low-code/no-code "automation application" provides the developer flexibility as the content grows and matures, and limits change the consumer may have to make in a later version.  As an example, #2 here may prove to be a fragile approach and more difficult to maintain long term and requires the consumer have specific details about the roles, files, and structure of the collection

```

- hosts: all
  gather_facts: false
  tasks:
  - name: Perform several actions
    include_role: mycollection.run
    vars:
      actions:
      - name: thing_1
        vars:
          thing_1_var_1: 1
          thing_1_var_2: 2
      - name: thing_2
        vars:
          thing_2_var_1: 1
          thing_2_var_2: 2

- hosts: all
  gather_facts: false
  tasks:
  - name: Do thing_1
    include_role: 
      name: mycollection.thing_1
      tasks_from: thing_1.yaml
      vars:
        thing_1_var_1: 1
        thing_1_var_2: 2
  - name: Do thing_2
    include_role: 
      name: mycollection.thing_2
      tasks_from: thing_2.yaml
      vars:
        thing_2_var_1: 1
        thing_2_var_2: 2
```
4) Provide sane defaults that work for most consumers, minimizing the number of required parameters required for use of the content. Provide configurable opt-in parameters when needed for flexibility.
5) Define the default values for user-configurable parameters in a roles' defaults. This is a familiar place to see and document why certain defaults were chosen and they play into the outcome of the automation.


I think some of these overlap with other areas, but this is what came to mind :)

jillr: Thanks @cidrblock.  I believe the doc currently covers the points about default, reworked the other suggestions into the doc format.

roles/README.adoc
jillr: Reworked this based on conversation at https://github.com/redhat-cop/automation-good-practices/pull/68/files#r932600169

roles/README.adoc
tima: FWIW Point 4 (and to an extent 5) in @cidrblock's response is convention over configuration design that I believe we need to apply a lot more than we have been to Ansible: https://en.wikipedia.org/wiki/Convention_over_configuration.

jillr:
@ericzolf I know I'm too late for the call today but when folks have a chance this should be ready for review

roles/README.adoc
```diff
@@ -19,6 +19,72 @@ The role internally would have the logic to decide whether to use ntpd, chronyd,
...
+      tasks_from: thing_2.yaml
+      vars:
+        thing_2_var_1: 1
+        thing_2_var_2: 2
+----
+
+====
+
+* Place content common to multiple roles in a single reusable role, "common" is a typical name for this role when roles are packaged in a collection.
+* Author loosely coupled hierarchical content.
```
ericzolf, line 83: Sorry if I missed this sentence during my first review but I simply don't understand it.

jillr: Added additional explanation text for this rec.

jillr:
Hi @ericzolf, would it be possible to get a re-review on this before the CoP meeting next week?

gravesm:
Why wouldn't you just use a simple variable structure? For example:
```yaml
- hosts: all
  gather_facts: false
  tasks:
  - name: Perform several actions
    include_role: mycollection.run
    vars:
      var_1: 1
      var_2: 2
      var_3: 3
      var_4: 4
```
If the point of the role is to hide the underlying implementation of the included roles, it seems like that should also extend to the variables. The example as it exists makes it look like the variables are just being passed through, which would expose the user to changes in those roles. You might need a bit more explanatory text since this example doesn't make it clear what the connection is to the underlying variables, but that kind of seems like what you'd want in this pattern.

jillr:
@ericzolf  Do you think gravesm's suggestion^ would be better?  Or should we just get rid of the contrived examples entirely?

coding_style/README.adoc
```diff
@@ -153,6 +153,22 @@ Rationale:: https://yaml.org/type/bool.html[YAML 1.1] allows all variants wherea
...
   In doubt, nobody expects JSON so stick to YAML.
+* Break up lengthy Jinja templates into multiple templates when there are distinct logical sections.
++
+[%collapsible]
+====
+Rationale:: Long and complex Jinja templates can be difficult to maintain and debug. By splitting excessively long templates into logical componets that can be included as-needed, each template will be easier to maintain.
+====
+
+* Jinja templates should not be used to create structured data but instead text and semi-structured data.
+* Filter plugins are preferred over Jinja templates for the use of data manipulation or transformation.
```
ericzolf, line 164:
```suggestion
* Jinja templates should not be used to create structured data but instead text and semi-structured data.
  Filter plugins are preferred over Jinja templates for the use of data manipulation or transformation.
```

As I understand it now, this is actually one statement with the same rationale, so I would merge both bullet points into one.

roles/README.adoc
```diff
@@ -19,6 +19,75 @@ The role internally would have the logic to decide whether to use ntpd, chronyd,
...
+      tasks_from: thing_2.yaml
+      vars:
+        thing_2_var_1: 1
+        thing_2_var_2: 2
+----
+
+====
+
+* Place content common to multiple roles in a single reusable role, "common" is a typical name for this role when roles are packaged in a collection.
+* Author loosely coupled, hierarchical content.
```
ericzolf, line 83:
```suggestion
  Author loosely coupled, hierarchical content.
```
Again, two phrases with the same explanation and rationale, so merge into one bullet.

ericzolf (changes requested):
OK, only two small formatting changes. If you can fix them until tomorrow 12EST, we can approve in the CoP meeting.

As a side note and warning, I had to update your branch with the latest 'main', so take care.

ericzolf:
I resolved all outdated comments, hopefully everything is clear now (only two small formatting fixes). Sorry for the big delay, this MR review got swamped in daily life. Thanks for supporting this effort!

ericzolf:
Approved yesterday in the CoP meeting, merging. Thanks for the effort (and the patience!).

Issue #17, opened 2021-05-20
Add GitHub Action to automatically generate the GitHub pages


ericzolf:
The file `docs/index.html` has currently to be generated manually, when required.

The following GitHub actions are required:

* make sure that each PR hasn't broken the asciidoc format 
* generate the pages on release version tags vX.Y.Z
* generate the pages to a non-too visible place e.g. docs/index.pre.html on pre-release tags (e.g. vX.Y.Z[ab|rc]N for alpha, beta, release candidate)
* generate the pages to a non-too visible place e.g. docs/index.dev.html on master branch

Overall:

* the tag/branch/commit should be in the generated HTML so that we can track what we see
* we might need to move the content of docs to a specific branch (if this is the decision, a project admin needs to change the configuration of GitHub pages, to take the pages from there instead of master).

ericzolf:
As examples, there are a few GitHub workflows on how to do different actions based on PR/branch/tag conditions: https://github.com/rdiff-backup/rdiff-backup/tree/master/.github/workflows

PR #71, opened 2022-06-23
Add automartion to build the docs

mophahr:
* on every push to a branch _other than_ main, previews are generated by an github action (can be seen as artefacts in the actions tab)
* every new tag on main that starts with `v` causes a new release of the rendered docs


mophahr:
closes #17

PR #75, opened 2022-07-18
updating wording in this copy/template section

djdanielsson:
I do not think it should be worded like this, there are many circumstances that you could know that you would not use a variable in a file and thus use copy.

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -293,7 +293,8 @@ Using these make code cleaner to read and ensure idempotence.
 ====
 
 * Limit use of the `copy` module to copying remote files, static files, and to uploading binary blobs.
-For most file pushes, use the `template` module. Even if there currently is nothing in the file that is being templated, if there is the possibility in the future that it might be added, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the copy and then needs to be moved before it can be edited.
+  For most file pushes, use the `template` module.
+  Even if there currently is nothing in the file that is being templated, if there is the possibility in the future that it might be added, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` module and then needs to be moved before it can be edited.
 * When using the `template` module, append `.j2` to the template file name.
 +
 [%collapsible]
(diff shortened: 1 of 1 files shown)


djdanielsson:
I am open to changing to wording more but I think it should be clear that if the user doesn't believe there is a case that templating would be used they SHOULD use copy.

coding_style/README.adoc
```diff
@@ -292,10 +292,9 @@ Alternatively, most configuration files have their own modules, such as https://
 Using these make code cleaner to read and ensure idempotence.
 ====
 
-* Limit use of the `copy` module to copying remote files and to uploading binary blobs.
-  For all other file pushes, use the `template` module.
```
djdanielsson, line 296: I think this line 296 is the one I would say I have the biggest problem with.

coding_style/README.adoc
```diff
@@ -292,10 +292,9 @@ Alternatively, most configuration files have their own modules, such as https://
...
-  Even if there is nothing in the file that is being templated at the current moment, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` and then needs to be moved before it can be edited.
-* When using the `template` module, append `.j2` to the template file name. 
+* Limit use of the `copy` module to copying remote files, static files, and to uploading binary blobs.
+For most file pushes, use the `template` module. Even if there currently is nothing in the file that is being templated, if there is the possibility in the future that it might be added, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the copy and then needs to be moved before it can be edited.
```
mophahr, line 296:
```suggestion
  For most file pushes, use the `template` module.
  Even if there currently is nothing in the file that is being templated, if there is the possibility in the future that it might be added, having the file handled by the `template` module now makes adding that functionality much simpler than if the file is initially handled by the `copy` module and then needs to be moved before it can be edited.
```

mophahr (changes requested):
one minor nitpick; other than that, this looks good :)

coding_style/README.adoc
djdanielsson: let me know if the current version fixed the issues you saw with the original

mophahr (approved):

sean-m-sullivan (approved):
LGTM

PR #86, opened 2023-02-08
Remove output folder from source

Fale:
The `docs` folder contains the HTML version of the document.

This folder should never be checked out (hence it should be in the `/gitignore` file).
Though, we need it to be present while creating the documentation (hence it should be created in the `release` and in the `preview` targets in the `Makefile`).
Also, the github release pipeline should create it (in the `docs` branch), hence the `--force` to the `git add` command in the release pipeline.


mophahr (approved):
Looks good

mophahr:
Approved in CoP meeting

PR #90, opened 2023-03-02
Use a Fedora container

Fale:
Since Ubuntu's repository seems to include a not always installable version of plantuml due to an issue with Java dependencies version, let's move to Fedora!


mophahr (approved):
LGTM

PR #93, opened 2023-03-22
Use OpenIconic

Fale:
Use OpenIconinc from Creole (https://plantuml.com/creole) instead of the symbol to ensure proper rendering.


mophahr (approved):
Approved in the CoP meeting

PR #94, opened 2023-03-29
Clarify use of templates in names

richm:
Templates used in names must come at the end of the name string.
See https://ansible-lint.readthedocs.io/rules/name - `name[template]`


djdanielsson (approved):
LGTM

PR #95, opened 2023-03-30
add more examples of line wrapping

richm:
Recommend the use of `>-` instead of `>` to avoid common problems.
Add a section at the end - "How to Wrap Lines"

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -63,14 +63,16 @@ example_list:
 +
 [%collapsible]
 ====
-Rationale:: long lines are difficult to read, many teams even ask for a line length limit around 120-150 characters.
-Examples:: there are multiple ways to avoid long lines but the most generic one is to use the YAML folding sign (`>`):
+Rationale:: Long lines are difficult to read, and many teams even ask for a line length limit around 120-150 characters.
+ansible-lint defaults to 82 characters per line - see (https://ansible-lint.readthedocs.io/rules/yaml/[Ansible Lint YAML rules]).
+You don't want to have to skip `yaml[line-length]` in your `.ansible-lint`, or litter your code with `# noqa yaml[line-length]`.
+Examples:: there are multiple ways to avoid long lines but the most generic one is to use the YAML folding sign (`>-`):
 +
 .Usage of the YAML folding sign
 [source,yaml]
 ----
 - name: call a very long command line
-  command: >
+  command: >-
     echo Lorem ipsum dolor sit amet, consectetur adipiscing elit.
     Maecenas mollis, ante in cursus congue, mauris orci tincidunt nulla,
     non gravida tortor mi non nunc.
@@ -82,7 +84,9 @@ Examples:: there are multiple ways to avoid long lines but the most generic one 
       Duis a neque fermentum nulla imperdiet blandit.
 ----
 +
-TIP: use the sign `>-` if it is important that the last line return code doesn't become part of the string (e.g. when defining a string variable).
+TIP: Always use the sign `>-` instead of `>` unless you are absolutely sure the trailing newline is not significant.
+It is really easy to introduce an error by using `>` and silently add a newline to a variable, like a filename, which leads to strange, hard to decipher errors.
+See below for "How To Wrap Lines"
 ====
 
 * If the `when:` condition results in a line that is too long, and is an `and` expression, then break it into a list of conditions.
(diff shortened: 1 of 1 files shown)


branic (approved):
LGTM

coding_style/README.adoc
```diff
@@ -82,7 +84,9 @@ Examples:: there are multiple ways to avoid long lines but the most generic one
       Duis a neque fermentum nulla imperdiet blandit.
 ----
 +
-TIP: use the sign `>-` if it is important that the last line return code doesn't become part of the string (e.g. when defining a string variable).
+TIP: Always use the sign `>-` instead of `>` unless you are absolutely sure the trailing newline is not significant.
```
mophahr, line 87:
```suggestion
TIP: Always use the sign `>-` instead of `>` unless you are absolutely sure the trailing newline is not significant.
The sing `>` adds a newline character to the last line, effecively turning `non nunc.` at the end of the example string above into  `"non nunc.\n"`,`>-` doesn't add the newline character.
```

coding_style/README.adoc
```diff
@@ -353,3 +357,114 @@ localhost | SUCCESS => {
...
 ====
+
+== How To Wrap Lines
+
+ansible-lint has a pretty short line length, which causes problems if you are trying to use good programming practices by having descriptive variable names, which usually end up being quite long.
+Here are some examples of how to deal with line wrapping in common scenarios:
+
+* Jinja expressions can be wrapped
+
+Within the Jinja expression, whitespace isn't significant, so take advantage of this to wrap lines as readable as possible:
```
mophahr, line 368:
```suggestion
Within the Jinja expression, whitespace and newline characters aren't significant, so take advantage of this to wrap lines into as readable a form as possible:
```

mophahr (changes requested):
I generally like this a lot and it really adds useful information to the AGPs, but I would prefer  if it would be turned into a format that fits the reference example in the Contribution Guide better: https://github.com/redhat-cop/automation-good-practices/blob/main/CONTRIBUTE.adoc#writing

McSim85:
Should we merge this, guys?

mophahr:
The section "== How To Wrap Lines" still does not look like the rest of our guide when rendered.

richm:
> The section "== How To Wrap Lines" still does not look like the rest of our guide when rendered.

Is the latest commit better?

ericzolf:
Let me be more prescriptive here because we're doing loops, and I'm very well aware that the guide could use some more consistency, not only here, and I'm asking to fix things you didn't break @richm, but we need to get better one step after the other:

1. the actual recommendation is  to wrap lines, hence it should be the title of the chapter (instead of "How to Wrap Lines", something like "Wrap longer lines of code").
2. then follows explanation and rationale, which would be a summary of what already today exists
3. and the 4 bullets are just different ways to apply the recommendation, and they all belong in the Examples section
4. `Use vars for locally scoped intermediate values` has nothing to do with wrapping lines and should be a different chapter, with its own title and Explanation/Rationale/Example sections

Does this make sense to everybody? I have more smaller comments, but that can wait until we've agreed on the overall structure.

I have more comments

richm:
> Let me be more prescriptive here because we're doing loops, and I'm very well aware that the guide could use some more consistency, not only here, and I'm asking to fix things you didn't break @richm, but we need to get better one step after the other:
> 
>     1. the actual recommendation is  to wrap lines, hence it should be the title of the chapter (instead of "How to Wrap Lines", something like "Wrap longer lines of code").
> 
>     2. then follows explanation and rationale, which would be a summary of what already today exists
> 
>     3. and the 4 (actually only 3, see next point) bullets are just different ways to apply the recommendation, and they all belong in the Examples section
> 
>     4. `Use vars for locally scoped intermediate values` has nothing to do with wrapping lines and should be a different chapter, with its own title and Explanation/Rationale/Example sections
> 
> 
> Does this make sense to everybody?

Yes.  Please see latest commit.

> I have more smaller comments, but that can wait until we've agreed on the overall structure.

coding_style/README.adoc
```diff
@@ -400,3 +405,91 @@ localhost | SUCCESS => {
...
+But what if the code is already indented a lot, and the variable I'm assigning to is already very long, and I can't put anything else on the line?
+Just start the assignment on the next line.
+
+* Use backslash escapes in double quoted strings
+
+But what if I have a very long string that I cannot use `>-` to wrap because I cannot have extra spaces in the value e.g. like a url value?
+Use a backslash escape in a double quoted string.
+YAML will concatenate the values with no spaces.
+
++
```
mophahr, line 430:
```suggestion
```

coding_style/README.adoc
```diff
@@ -400,3 +405,91 @@ localhost | SUCCESS => {
...
+
+                  - name: Wrap when first line is already too long
+                    very_indented_foo: "{{
+                      a_very.long_variable.name |
+                      somefilter('with', 'many', 'arguments') |
+                      another_filter | list }}"
+                    when: \
+                      a_very.long_variable.name |
+                      somefilter('with', 'many', 'arguments') |
+                      another_filter | list | length > 0
```
mophahr, line 454:
```suggestion
- name: Wrap when first line is already too long
  very_indented_foo: "{{
    a_very.long_variable.name |
    somefilter('with', 'many', 'arguments') |
    another_filter | list }}"
  when: \
    a_very.long_variable.name |
    somefilter('with', 'many', 'arguments') |
    another_filter | list | length > 0
```

coding_style/README.adoc
```diff
@@ -400,3 +405,91 @@ localhost | SUCCESS => {
...
+- name: Very long line with Jinja expression
+  foo: "{{ a_very.long_variable.name | somefilter('with', 'many', 'arguments') | another_filter | list }}"
+
+                  - name: First line is already too long
+                    very_indented_foo: "{{ a_very.long_variable.name |
+                      somefilter('with', 'many', 'arguments') |
+                      another_filter | list }}"
+                    when: a_very.long_variable.name |
+                      somefilter('with', 'many', 'arguments') |
+                      another_filter | list | length > 0
```
mophahr, line 484:
```suggestion
- name: First line is already too long
  very_indented_foo: "{{ a_very.long_variable.name |
    somefilter('with', 'many', 'arguments') |
    another_filter | list }}"
  when: a_very.long_variable.name |
    somefilter('with', 'many', 'arguments') |
    another_filter | list | length > 0
```

coding_style/README.adoc
```diff
@@ -400,3 +405,91 @@ localhost | SUCCESS => {
...
+
+== Wrap longer lines of code
+
+ansible-lint has a pretty short line length, which causes problems if you are trying to use good programming practices by having descriptive variable names, which usually end up being quite long.
+Here are some examples of how to deal with line wrapping in common scenarios:
+
+* Jinja expressions can be wrapped
+
+Within the Jinja expression, whitespace and newline characters aren't significant, so take advantage of this to wrap lines into as readable a form as possible.
+Remember, in a `when`, `that`, `failed_when`, or other such keywords, you can just write Jinja code - you do not need the `"{{ ... }}"`
```
mophahr, line 417:
```suggestion
* Jinja expressions can be wrapped.
  Within the Jinja expression, whitespace and newline characters aren't significant, so take advantage of this to wrap lines into as readable a form as possible.
  Remember, in a `when`, `that`, `failed_when`, or other such keywords, you can just write Jinja code - you do not need the `"{{ ... }}"`
```

coding_style/README.adoc
```diff
@@ -400,3 +405,91 @@ localhost | SUCCESS => {
...
+* Start an expression with '{{' followed by newline if the line will otherwise be too long
+
+But what if the code is already indented a lot, and the variable I'm assigning to is already very long, and I can't put anything else on the line?
+Just start the assignment on the next line.
+
+* Use backslash escapes in double quoted strings
+
+But what if I have a very long string that I cannot use `>-` to wrap because I cannot have extra spaces in the value e.g. like a url value?
+Use a backslash escape in a double quoted string.
+YAML will concatenate the values with no spaces.
```
mophahr, line 428:
```suggestion
* Start an expression with '{{' followed by newline if the line will otherwise be too long.
  But what if the code is already indented a lot, and the variable I'm assigning to is already very long, and I can't put anything else on the line?
  Just start the assignment on the next line.

* Use backslash escapes in double quoted strings.
  But what if I have a very long string that I cannot use `>-` to wrap because I cannot have extra spaces in the value e.g. like a url value?
  Use a backslash escape in a double quoted string.
  YAML will concatenate the values with no spaces.
```

mophahr (changes requested):
Thanks @richm 

I have made a few suggestions to align the formatting with the rest of the doc. When they are applied, there is no further objection from my side to merge this.

mophahr:
Looks good now. 

As we have discussed in today's CoP meeting, this is ready to be merged.

mophahr (approved):

PR #111, opened 2024-04-11
Examples pass ansible-lint.

jeffmcutter:


inventories/inventory_loop_hosts/playbook_bad.yml
```diff
@@ -1,12 +1,14 @@
-- name: provision hosts in a bad way
+- name: Provision hosts in a bad way
```
djdanielsson, line 1: I think this playbook is supposed to be bad so maybe it shouldn't pass lint?

djdanielsson (changes requested):
I did a quick skim and I think at least this one playbook is supposed to not be the way to do it so I am not sure if we care or want it to pass lint? Just a thought, I am not sure if there are any others like this it was the only one I noticed.

inventories/inventory_loop_hosts/playbook_bad.yml
jeffmcutter: I considered that as well, but was thinking the focus was more around the approach to the problem than about the formatting of the file.  I can change it back if we want.

djdanielsson: @ericzolf what is your opinion here?

ericzolf: I agree with Jeff, let's keep it 100% clean unless we have a very good reason (or the next person going through this exercise will trip over it).

ericzolf (approved):

djdanielsson (approved):
LGTM

PR #117, opened 2024-12-05
Add html link checker schedule

### LINKED ISSUE #99: Add link-checking to the CiCD pipeline

mophahr:
closes #99

Together with the markdown link check we now can be sure we don't have broken links in the code nor in the published page. The new action runs on a schedule so that we get notified even if we don't make any changes.

To see it in action have a look at https://github.com/mophahr/test_link_checker


djdanielsson (approved):
LGTM

PR #118, opened 2025-02-08
Fix missing period and duplicate 'file' word

cavcrosby:


ericzolf:
Without further ado, thanks for the fix, merging!

PR #119, opened 2025-03-07
Add recommendation to use symbolic file permissions.

jkupferer:
As explained in the text and recommended in the Red Hat training materials, symbolic file permissions are easier to understand and less prone to mistakes. Particularly with YAML octal notation, symbolic mode should be the standard.

### DIFF
--- a/coding_style/README.adoc
+++ b/coding_style/README.adoc
@@ -400,3 +400,34 @@ localhost | SUCCESS => {
 
 ```
 ====
+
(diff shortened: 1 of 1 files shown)


djdanielsson:
I disagree with this, I do not think we should be pushing symbolic over octal

ericzolf:
We discussed this topic recently in the Automation CoP and there was a rather strong opposition to this recommendation, most people felt that octal is still the way to go. So I will close this pull request accordingly, sorry, don't hesitate to suggest further ideas.

PR #123, opened 2025-04-04
fixing bad link

djdanielsson:
latest was removed it would appear so currently setting it to 4.4, likely there is an AAP doc page that is similar that we could maybe update later but this will stop the GitHub action making issues about his constantly.

#122 #121


mophahr:
thanks

mophahr (approved):

PR #126, opened 2025-04-14
Add a rule around inventory and credential matching

bontreger:
This topic has come up a few times in the past week around managing multiple credentials on a single job template. Adding this rule to be explicit on one-to-one inventory/credential/job template mapping

### DIFF
--- a/inventories/README.adoc
+++ b/inventories/README.adoc
@@ -24,6 +24,21 @@ Ansible provides a lot of https://docs.ansible.com/ansible/latest/plugins/invent
 This gives you a complete model of the environment to be automated, with limited effort to maintain it, and no confusion about where to modify it to get the result you need.
 ====
 
+== An inventory should use a single machine credential
+[%collapsible]
+====
+Explanations::
+An inventory should use a single machine (connection) credential. Inventories should not be mixed with hosts that use different credentials.
+
+Rationale::
+When running a playbook, only a single machine credential can be specified within AWX/Ansible Automation Platform.  Additionally, a single password or SSH key can be provided if operating from the command line.
+
+Examples::
+A typical enterprise implementation may have hosts of the same type, for instance, Linux hosts, with different credentials for different environments, such as in a DMZ or a development environment.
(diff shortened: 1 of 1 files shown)


djdanielsson (commented):
I think we need to probably create a new section for AAP related practices.

bontreger:
I think inventory/credential mapping still makes sense in a CLI implementation only manner.  You can get clever with defining connection details in an inventory, and while possible, it's a slippery slope and unmaintainable long-term

ericzolf:
I don't agree with the way this is formulated: it would be for me completely valid to have Linux and Windows hosts in the same inventory, and they would have different machine credentials, albeit at the group level probably. Also the fact that you express the recommendation with a "should" is in my eyes a bad sign, that means you're not sure it applies always.

Shouldn't the recommendation actually be to not use _individual_ credentials? This can be easily argumented against, due to maintenance effort _or_ when using an external vault, performance impact because you'd need to query its API individually for each host (a colleague heard this actually from a vault provider employee, can't remember which one).

And this recommendation is independent from the execution place, AAP or CLI.

inventories/README.adoc
```diff
@@ -24,6 +24,21 @@ Ansible provides a lot of https://docs.ansible.com/ansible/latest/plugins/invent
...
 ====
 
+== An inventory should use a single machine credential
+[%collapsible]
+====
+Explanations::
+An inventory should use a single machine (connection) credential. Inventories should not be mixed with hosts that use different credentials.
+
+Rationale::
+When running a playbook, only a single machine credential can be specified within AWX/Ansible Automation Platform.  Additionally, a single password or SSH key can be provided if operating from the command line.
```
mophahr, line 34:
```suggestion
When running a playbook, only a single machine credential can be specified within AWX/Ansible Automation Platform. Additionally, a single password or SSH key can be provided if operating from the command line.
```

mophahr (commented):
I like it, but I have one tiny suggestion

mophahr:
@bontreger @djdanielsson what's the plan with this PR?

djdanielsson:
At a min it needs to be rewritten/rebased to the new layout but I would have to look into the actual topic to see if I agree or not

PR #154, opened 2025-05-20
Exclude all of github from link checker

mophahr:
to resolve false-positive missing link issues caused by github replying with HTTP_429


djdanielsson (approved):
LGTM

PR #173, opened 2026-05-19
docs(roles): require loop_control.loop_var with __ prefix

ivarmu:
Document that role loops must not use bare item; use internal __rolename_* loop variables. Update platform vars examples.

### DIFF
--- a/roles/README.adoc
+++ b/roles/README.adoc
@@ -124,6 +124,25 @@ Using the name of the role reduces the potential for name conflicts and using th
 The two underscores convention has prior art in some popular roles like
 https://github.com/geerlingguy/ansible-role-apache/blob/f2b91ac84001db3fd4b43306a8f73f1a54f96f7d/vars/Debian.yml#L8[geerlingguy.ansible-role-apache]).
 This includes variables set by set_fact and register, because they persist in the namespace after the role has finished!
+* In roles, do not use the default Ansible loop variable `item`. Set `loop_control.loop_var` to an internal name prefixed with `__` (and ideally the role name), then reference that name in the task `vars:` block and in `when:` conditions.
++
+Rationale:: `item` is generic and leaks into the play namespace; it is easy to clash with another loop or with a host variable. A dedicated `__rolename_*` loop variable makes intent obvious and matches the internal-variable convention.
++
+Example::
++
+[source,yaml]
+----
+- name: Set platform/version specific variables
+  ansible.builtin.include_vars: "{{ __rolename_vars_file }}"
+  loop:
+    - "{{ ansible_facts['os_family'] }}.yml"
+    - "{{ ansible_facts['distribution'] }}.yml"
+  loop_control:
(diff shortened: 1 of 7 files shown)


djdanielsson (approved):
LGTM

lennysh (approved):
LGTM

roles/dont_use_groups/playbook.yml
```diff
@@ -6,13 +6,17 @@
...
-        msg: do something with {{ item }}
+        msg: do something with {{ __cluster_member }}
       loop: "{{ groups[cluster_group_name] }}"
+      loop_control:
+        loop_var: __cluster_member
 
     - name: The loop happens only for the first host in each group
       ansible.builtin.debug:
-        msg: do something with {{ item }}
+        msg: do something with {{ __cluster_member }}
```
jeffcpullen, line 16: I believe this would need to be quoted

roles/dont_use_groups/playbook.yml
```diff
@@ -21,8 +25,10 @@
 
     - name: The loop happens only for the first _available_ host in each group
       ansible.builtin.debug:
-        msg: do something with {{ item }}
+        msg: do something with {{ __cluster_member }}
```
jeffcpullen, line 28: Same as above, quotations likely needed to pass linter

roles/dont_use_groups/playbook2.yml
```diff
@@ -19,5 +21,7 @@
   tasks:
     - name: The loop happens for each execution host
       ansible.builtin.debug:
-        msg: do something with {{ item }}
+        msg: do something with {{ __cluster_member }}
```
jeffcpullen, line 24: Same as above, likely needs quotes

roles/README.adoc
```diff
@@ -124,6 +124,25 @@ Using the name of the role reduces the potential for name conflicts and using th
 The two underscores convention has prior art in some popular roles like
 https://github.com/geerlingguy/ansible-role-apache/blob/f2b91ac84001db3fd4b43306a8f73f1a54f96f7d/vars/Debian.yml#L8[geerlingguy.ansible-role-apache]).
 This includes variables set by set_fact and register, because they persist in the namespace after the role has finished!
+* In roles, do not use the default Ansible loop variable `item`. Set `loop_control.loop_var` to an internal name prefixed with `__` (and ideally the role name), then reference that name in the task `vars:` block and in `when:` conditions.
```
jeffcpullen, line 127: avoid dot notation is another good practice -> loop_control.loop_var

jeffcpullen (changes requested):
A few minor issues

roles/README.adoc
jeffcpullen: Actually, I read this as a literal variable but you are actually showing the parameter path. So its not really a dot notation issue, but rather a bit vague in what you are referencing. If someone wasn't familiar with the loop_var parameter, would this make sense to them?

PR #187, opened 2026-07-13
docs(coding_style): add !unsafe guidelines section

### LINKED ISSUE #109: Add use of !unsafe to good practices
It was a long standing idea, which I've moved now from the CoP minutes to this issue.

gferreir:
## Summary

Adds a new guideline entry under the **YAML and Jinja2 Syntax** section of
`coding_style/README.adoc` on the use of the `!unsafe` tag, addressing #109.

## What was added

A new bullet point with a collapsible block containing:

- **Explanation** — describes how `!unsafe` prevents Jinja2 from evaluating
  strings that contain `{` or `}` characters
- **Rationale** — covers the failure mode caused by unintended template
  evaluation and why `!unsafe` is preferred over `{% raw %}` blocks
- **Do / Don't examples** — YAML examples with regex and template-like strings
  using `!unsafe` vs bare strings that would cause Jinja2 errors

Closes #109


coding_style/README.adoc
```diff
@@ -183,6 +183,34 @@ Custom filter plugins can be written to handle complex or unique use-cases.
...
+Rationale:: Without `!unsafe`, Ansible will attempt to evaluate any string containing `{{` as a Jinja2 template expression.
+This causes failures or unexpected substitutions when the string is intentionally literal.
+Using `!unsafe` is cleaner and more explicit than escaping individual braces with `{% raw %}` blocks.
+
+.Do this:
+[source,yaml]
+----
+vars:
+  prometheus_query: !unsafe "rate(http_requests_total{job='api'}[5m])"
+  grafana_template: !unsafe "{{ server }}"
```
mophahr, line 207: wouldn't this work just as well whilst still looking more like "normal" Ansible? :
`grafana_template: "{{ '{{ server }}' }}"`
The !unsafe string probalby works more easily with more complicated expression.
Probably it's a matter of taste but maybe we should recommend both of them?

gferreir: Good point! Both work, but there's no clear guidance yet on when to prefer one over the other.

For simple cases like `{{ server }}` the Jinja2 escaping is a fine alternative. The problem starts when expressions get more complex — like regex patterns or config strings from other tools with multiple `{` and `}`. At that point each `{{` needs to be escaped individually, which gets messy fast.

The official docs also mention that `!unsafe` is more comprehensive than Jinja2 escaping since it marks the value at the data type level, not just at render time.

Maybe we could document both and give some guidance on when each fits better?

mophahr: > Maybe we could document both and give some guidance on when each fits better?

that would be my preferred solution

gferreir:
Updated! The section now documents both approaches:

- For simple single-expression cases, Jinja2 quoting (`{{ '{{ server }}' }}`) is shown as a valid alternative alongside `!unsafe`.
- The rationale now explains when to prefer each: `!unsafe` is the cleaner option for strings with multiple brace sequences (like regex patterns), where Jinja2 quoting would require escaping each `{{` individually.
- Added a Don't-do-this entry for `grafana_template` — without `!unsafe`, Ansible silently resolves the variable instead of passing the literal string.

mophahr (approved):

PR #188, opened 2026-07-14
docs(execution_environments): add best practices for building and managing EEs

### LINKED ISSUE #120: Add best practices for Execution Environments
Some guidance on the creation and use of EEs would be nice. [This blog](https://www.redhat.com/en/blog/ansible-execution-environment-tips) from @sean-m-sullivan is a great start.

gferreir:
Closes #120

This PR adds a new `execution_environments/` section with practical guidance on
building and managing Ansible Execution Environments, addressing the request in
issue #120.

## What was added

A new `execution_environments/README.adoc` file with 7 guidelines, each
following the established `[%collapsible]` + `Explanations` + `Rationale` +
examples pattern used throughout the document:

1. **Always use `version: 3`** in `execution-environment.yml` — required for
   `ansible-builder` 3.x; omitting it either causes an immediate `DefinitionError`
   (when using v3 syntax) or silently produces a limited v1 image (when using
   legacy syntax).

2. **Use `ee-minimal` as the base image** for custom EEs instead of
   `ee-supported`, which bundles pre-installed collections that may conflict
   with the versions the automation actually requires.

3. **Never use `:latest`** on the base image — the `:latest` tag was removed
   from official Red Hat `ee-minimal` images starting with AAP 2.7. Documents
   the two current naming schemes (versioned and version-less) and the SHA
   digest option for maximum reproducibility.

4. **Do not create monolithic EEs** — group by automation domain (Linux,
   networking, Windows, cloud) to avoid dependency conflicts and image bloat.

5. **Pin all dependency versions** — collections, Python packages, and system
   packages. Without pinning, two builds from the same definition file at
   different times can produce different images.

6. **Set `package_manager_path`** when using minimal base images — `ee-minimal`
   ships with `microdnf`, not `dnf`; omitting this option causes build failures
   at the system package installation stage.

7. **Store EE definitions in Git and automate builds** through CI/CD pipelines,
   with a reference to the `infra.ee_utilities` collection from
   `redhat-cop/ee_utilities` for integrating builds with Automation Controller.

## Notes

- The section is included in Part III (AAP and Configuration as Code) of the
  root `README.adoc`, alongside the existing `aap_configuration` section.
- Cross-references were added to the `aap_configuration`, `git_workflow`, and
  `security` sections where related guidance already exists.


mophahr (approved):
Nice addition to the guide. I'd like to get some more reviews if possible before we merge.

tech2734 (approved):
nice addition

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+
+For good practices on declaring EEs as objects in Automation Controller configuration, see <<aap_configuration/README.adoc#_aap_configuration_as_code,AAP Configuration as Code>>.
+For good practices on immutable references to EE images in production, see <<git_workflow/README.adoc#_never_use_latest_or_branch_names_in_production_references,Git Workflow: Never use "latest" or branch names in production references>>.
+For vulnerability scanning of built EE images, see <<security/README.adoc#_scan_container_images_and_generate_sboms_for_every_release,Security: Scan container images and generate SBOMs for every release>>.
+
+== Always use `version: 3` in the `execution-environment.yml` definition file
+[%collapsible]
+====
+Explanations::
+The `execution-environment.yml` definition file must specify `version: 3` as its first key.
```
sabre1041: YAML is unordered so there is no explicit requirement that `version: 3` must be the first key in the file

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+[%collapsible]
+====
+Explanations::
+Red Hat provides two categories of supported base images:
++
+* `ee-minimal` -- contains only `ansible-core` and `ansible-runner`.
+It is the recommended starting point for custom EEs.
+* `ee-supported` -- includes a curated set of pre-installed certified collections.
+It is designed to be used as-is in environments where its bundled content is sufficient.
+Using it as a base for custom images and then specifying additional or different versions of the same collections creates conflicts that are difficult to diagnose.
```
sabre1041:
```suggestion
Using it as a base for custom images and then specifying additional or different versions of the same collections often creates conflicts that are difficult to diagnose.
```

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+      - name: amazon.aws
+        version: "8.0.0"
+----
+====
+
+== Always pin the base image to a specific version -- never use `:latest`
+[%collapsible]
+====
+Explanations::
+The base image tag in the `images.base_image.name` field must always reference a specific version, not the `:latest` shorthand.
```
sabre1041:
```suggestion
The base image tag in the `images.base_image.name` field must always reference a specific tag version, and must not use `:latest` as a value.
```

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
++
+Red Hat distributes EE base images under two naming schemes, both requiring an explicit version:
++
+[cols="1,2,2"]
+|===
+| Type | Example | Use case
+
+| *Versioned* (tied to an AAP release)
+| `registry.redhat.io/ansible-automation-platform-27/ee-minimal-rhel9:2.7.0`
```
sabre1041: This is not resolvable

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+Each EE should contain only the dependencies required by a specific domain of automation.
+Common domain groupings include:
++
+* Linux / RHEL system management
+* Windows management
+* Networking (Cisco, Juniper, Arista, etc.)
+* Cloud (AWS, Azure, GCP)
+* Virtualization (VMware, OpenShift Virt)
++
+A separate, lightweight EE should be created for each domain rather than a single large image containing all collections.
```
sabre1041:
```suggestion
A separate, lightweight EE should be created for each domain rather than a single large image containing all collections and dependencies.
```

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+
+Rationale::
+Combining unrelated automation workloads into a single EE creates several problems:
++
+* *Dependency conflicts* -- networking collections may require specific Python library versions that conflict with those required by cloud collections.
+* *Image bloat* -- unused collections increase image size, slowing pulls and increasing storage costs in the registry.
+* *Maintenance burden* -- a change in any dependency forces a rebuild and validation of the entire image, even for unrelated automation.
```
sabre1041:
```suggestion
* *Maintenance burden* -- a change in any dependency forces a rebuild and revalidation of the entire image, even for unrelated automation.
```

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+====
+Explanations::
+Every dependency referenced in an EE definition must be pinned to an exact version:
++
+* Collections in `dependencies.galaxy`: use the `version:` key with an exact version string (e.g., `"1.23.0"`).
+* Python packages in `dependencies.python`: use `==` pinning (e.g., `jmespath==1.0.1`).
+* System packages in `dependencies.system`: use the specific package name; avoid wildcards.
+* `ansible-core` and `ansible-runner` in `dependencies.ansible_core` / `dependencies.ansible_runner`: pin with `package_pip: ansible-core==<version>`.
++
+An external `requirements.yml` or `requirements.txt` file is equally valid, but the same pinning requirement applies to those files.
```
sabre1041, line 311: Any guidance between defining dependencies inline instead of in external files

execution_environments/README.adoc
```diff
@@ -0,0 +1,528 @@
...
+----
+====
+
+== Store EE definitions in Git and automate builds through a CI/CD pipeline
+[%collapsible]
+====
+Explanations::
+Every `execution-environment.yml` file, along with any supporting `requirements.yml`, `requirements.txt`, and `bindep.txt` files, must be stored in a Git repository.
++
+Builds must be triggered automatically by CI/CD pipelines on relevant Git events (push to a watched branch, pull request merge, or a scheduled trigger for regular security updates).
```
sabre1041:
```suggestion
Builds must be triggered automatically by CI/CD pipelines on relevant Git events (push to a watched branch, pull requests, or a scheduled trigger for regular security updates).
```

sabre1041 (commented):
Looks good a few comments/suggestions

Any thoughts about including details on how to inject details related to retrieving collections from Automation Hub?

execution_environments/README.adoc
gferreir, line 311: Good question — we intentionally left this as 'either is valid' without a strong preference, but it may be worth adding explicit guidance. A few trade-offs worth discussing:

**Inline (inside `execution-environment.yml`):**
- Single file to track; easier to review as a unit
- Ideal for simple EEs with a small number of dependencies
- Version pins are immediately visible alongside the image and options config

**External files (`requirements.yml`, `requirements.txt`, `bindep.txt`):**
- Allows reuse across multiple EE definitions that share a common dependency set
- More familiar to users coming from classic Ansible workflows (requirements.yml is already a known pattern)
- Easier to integrate with tooling that expects separate requirements files

A reasonable recommendation might be: **prefer inline for new EEs** (keeps everything in one place, reduces file count), and **use external files when dependencies are shared across multiple EE definitions** or when integrating with existing tooling that expects them separately.

Happy to add a short note to this effect if the maintainers agree this guidance is useful here.

gferreir:
Thanks for the question about Automation Hub authentication — this is a real gap in the current content.

We have a brief note at the end of the CI/CD practice mentioning that `redhat.*` collections require Automation Hub credentials, but no practical example of how to actually pass them during a build.

The documented approach using `additional_build_steps.prepend_galaxy` with `ENV` + `ARG` directives is the right pattern:

```yaml
additional_build_steps:
  prepend_galaxy:
    - ENV ANSIBLE_GALAXY_SERVER_LIST=automation_hub
    - ENV ANSIBLE_GALAXY_SERVER_AUTOMATION_HUB_URL=https://console.redhat.com/api/automation-hub/content/<synclist>/
    - ENV ANSIBLE_GALAXY_SERVER_AUTOMATION_HUB_AUTH_URL=https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token
    - ARG ANSIBLE_GALAXY_SERVER_AUTOMATION_HUB_TOKEN
```

The `ARG` directive ensures the token is available during the galaxy install step but is not persisted in the final image metadata — an important security distinction from `ENV`.

Before adding this to the document, I'd like to understand the intended scope a bit better:
- Would you prefer this as an expansion within the existing CI/CD practice (Practice 7), or as a standalone practice?
- Should it also cover the `ansible.cfg` approach (configuring `[galaxy_server.automation_hub]` and pointing to it) as an alternative?

Happy to draft either approach — just want to make sure it fits the structure the maintainers have in mind before writing it.

jeffcpullen:
Thanks for picking this up, the build-time credential gap is worth closing.

In regards to the galaxy/hub authentication I'd lean standalone rather than folding it into the CI/CD practice. Passing a secret into a container build safely is a container-build concern, not a CI/CD one (it applies just as much to a developer building an EE locally), so it reads better as its own good practice that the CI/CD section can link to.

The good practice here is isn't really `ARG` vs `ENV`, it's that `ansible-builder` generates a multi-stage build and runs the galaxy install in an intermediate stage that gets discarded, so the final image copies only the installed collections forward. `ARG` on its own isn't secret-safe: in a single-stage build an `ARG` value is recorded in the layer's build history and shows up in `docker history` (`|1 TOKEN=... /bin/sh -c ...`). So a reader who isn't using a multi-stage build tool like `ansible-builder` takes away "use `ARG`, not `ENV`" could apply it to a hand-written single-stage `Containerfile` and leak the token.

The tool-agnostic durable principal is **a build-time secret must never persist in the shipped image.** Keep it to the step that needs it (a discarded build stage, or a `RUN --mount=type=secret` mount), and verify it's gone with `docker history` / `skopeo inspect` before pushing. `ENV` is then correctly the tool for the non-secret bits (the server list and URLs), and the token rides an `ARG` precisely because `ansible-builder`'s galaxy stage is discarded. Worth saying that explicitly, so the safety is attributed to the stage boundary rather than the keyword.

On the `ansible.cfg` alternative: worth a mention, with the same caveat, that the credential must not end up written into a file that gets copied into the image. Supplied at build time and not committed or copied in, it's fine.

jeffcpullen:
Useful addition (per #120). An issue I see is that it's pitched to the current tooling (the inline `version: 3` schema, exact image names, the "removed in AAP 2.7" fact, the `microdnf` path, a pinned `ansible-builder`), where the rest of GPA states durable principles and defers the specifics. Each would age better and see less churn by focusing on the principles, with the details left to the upstream docs the section already links. 

- **`version: 3`**  -> declare the schema version you build against rather than relying on the default
- **`ee-minimal` base** -> build up from a minimal base rather than inheriting collections you didn't declare (holds on community and OS bases too, not just entitled registries).
- **never `:latest`** -> no need to re-pin it to the AAP 2.7 catalog change.
- **`package_manager_path`** -> match the base image's package manager, not the specific `microdnf` path.
- **Git + CI/CD** -> version the definitions and build them in CI

"Pin all dependency versions" and "Do not create monolithic EEs" already sit about right.

gferreir:
Thanks for laying that out, @jeffcpullen — the distinction between principle and tooling detail makes sense, and looking at the rest of the guide through that lens I can see the current draft leans too heavily on specifics.

The Automation Hub point is fair too. "Use `ARG`, not `ENV`" is accurate for ansible-builder but it's the wrong takeaway if someone applies it to a hand-written single-stage build. The principle you described — credential must not persist in the shipped image — is the right anchor, and the `ARG` + discarded stage is just how ansible-builder happens to implement it today.

So the direction I'm hearing is: keep the practices, but pull the prose back to the "why" and let the upstream docs carry the "how". That makes sense for longevity.

One thing I'm genuinely unsure about: should the Do/Don't examples stay? My gut says yes — a reader who's never seen an `execution-environment.yml` before needs something concrete to orient around. But I could see those examples being the thing that anchors the text to a specific version of the tooling.

What if the examples stayed but the explanatory text made clear they're illustrating the principle, not prescribing the exact syntax? Something like:

> **Declare the schema version explicitly**
> Always declare the `version` key. Without it, the builder relies on a default that may change between releases.
> *(The current schema version is `3` — see the [ansible-builder docs](link) for what's supported.)*

That way the principle doesn't expire when the version bumps. If the consensus is to drop the specifics entirely and rely on external links, I can do that too — just want to align before rewriting.

jeffcpullen:
> One thing I'm genuinely unsure about: should the Do/Don't examples stay? My gut says yes — a reader who's never seen an `execution-environment.yml` before needs something concrete to orient around. But I could see those examples being the thing that anchors the text to a specific version of the tooling.

I would say it depends. Examples are always helpful and give some more freedom to get specific, but there is no mechanism to identify drift between the source of truth and here other than users opening up issues that something doesn't work the way it says it does. So try to avoid examples that will rot unnoticed. 

To give an example, the ansible-builder schema has evolved over time and I'm not sure if it has reached a point where we can expect that version for a long time (I haven't dug into this, just guessing). Where as the Dockerfile/Containerfile standards are pretty mature at this point and the features that users need are very stable. So giving an example of a multi-stage build Containerfile that takes in ARGs for the configurable bits like the image, packages, secrets, etc is unlikely to change. The value of ansible-builder is that it abstracts a lot of those details and is an easier way to adhere to these good practices.

> What if the examples stayed but the explanatory text made clear they're illustrating the principle, not prescribing the exact syntax? Something like:
> 
> > **Declare the schema version explicitly**
> > Always declare the `version` key. Without it, the builder relies on a default that may change between releases.
> > _(The current schema version is `3` — see the [ansible-builder docs](link) for what's supported.)_
> 
> That way the principle doesn't expire when the version bumps. If the consensus is to drop the specifics entirely and rely on external links, I can do that too — just want to align before rewriting.

This seems like the right way to handle it.

gferreir:
Pushed a new commit based on this discussion — here's what it covers.

**Principle-first reframe (per @jeffcpullen's feedback)**

Reworked the prose in sections 1, 2, 3, 6, and 7 to lead with the durable principle and move tooling specifics to callouts or notes with links to upstream docs:

- "Declare the schema version explicitly" (was "Always use `version: 3`") — the current version is now a callout note, not part of the principle statement
- "Build from a minimal base — do not inherit collections you did not declare" (was "Use `ee-minimal`") — principle now holds for any base image, not just Red Hat's
- `:latest` rationale no longer references the AAP 2.7 deprecation — reproducibility is the argument, product version is not
- `package_manager_path` heading now reads "match the base image's package manager" — `/usr/bin/microdnf` is an annotation, not the principle
- CI/CD example no longer hard-codes the `ansible-builder` version — a callout points to PyPI and says to pin it as you would any build dependency

The Do/Don't examples stayed. Following the direction from this thread: examples help readers orient, but the explanatory text now makes clear they illustrate the principle rather than prescribe the exact syntax.

**New practice: Never persist build-time credentials in the shipped image**

Added as a standalone section addressing the Automation Hub question from @sabre1041. The principle is the stage boundary, as @jeffcpullen described — `ARG` is safe here because `ansible-builder` generates a multi-stage build and discards the galaxy stage. The section includes a `CAUTION` note clarifying that `ARG` alone is not secret-safe in single-stage builds, to avoid the misreading we discussed.

Happy to iterate from here if anything still needs adjusting.

myllynen:
Given that this document targets AAP and already recommends using the ee-minimal image which requires AAP subscription, perhaps the CI/CD example should use the readily available UBI-based ansible-builder image instead of using a non-supported base image and then installing something from upstream?

Derek pointed out that it might be ideal here, for the ansible-builder image details see:

https://catalog.redhat.com/en/software/containers/ansible-automation-platform-27/ansible-builder-rhel9/69fb1e2545f6af701dad8723#containerfile

gferreir:
Agreed. Replaced the upstream `pip install ansible-builder` step with
the `ansible-builder-rhel9` container image from the Red Hat Ecosystem
Catalog.

The CI/CD example now uses a two-step approach:

1. `ansible-builder create` runs inside the Red Hat image to generate
   the build context — no container runtime needed inside the image.
2. `docker build` runs on the CI runner against the generated
   `Containerfile`.

This also makes `registry.redhat.io` authentication explicit in the
workflow, which was already required for pulling `ee-minimal` but was
not shown before.

mophahr:
Hi @gferreir @jeffcpullen @myllynen @sabre1041 

Thank you very much for the PR and the reviews/contributions!

From teh comment history here, I'm unable to assess waht the state of this PR is. Apart from teh fact, that it needs rebasing, what review comments are still open as of now? Or in other words: what needs to be changed to get all reviewers to approve.

jeffcpullen:
@mophahr My initial feedback around the changes being overly specific and prone to going stale have been addressed. I'm good with it.

gferreir:
Hi @mophahr — I believe all review comments have been addressed and responded to in the thread. I'll rebase against main now. Let me know if there's anything else needed from my side.

